PR #6, opened 2024-06-13
Initial skeleton for Evaluator classes and exceptions

nathan-weinberg:
This PR introduces the `Evaluator` parent class and the `MMLUEvaluator` and `MTBenchEvaluator` child classes

It also introduces the `EvalError` parent exception class for all `eval` custom exceptions and the `ModelNotFoundError` child exception class

### DIFF
--- a/src/instructlab/eval/evaluator.py
+++ b/src/instructlab/eval/evaluator.py
@@ -11,6 +11,3 @@ class Evaluator:
 
     def __init__(self, model: str) -> None:
         self.model = model
-
-    def run(self) -> dict:
-        return {}
--- a/src/instructlab/eval/evaluator.py
+++ b/src/instructlab/eval/evaluator.py
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+class Evaluator:
+    """
+    Parent class for Evaluators
+    """
+
+    def __init__(self, model) -> None:
+        self.model = model
+
+    def run(self) -> dict:
+        return {}
--- a/src/instructlab/eval/evaluator.py
+++ b/src/instructlab/eval/evaluator.py
@@ -4,9 +4,12 @@
 class Evaluator:
     """
     Parent class for Evaluators
+
+    Atttributes:
+        model   The model to be evaluated
     """
 
-    def __init__(self, model) -> None:
+    def __init__(self, model: str) -> None:
         self.model = model
 
     def run(self) -> dict:
--- a/src/instructlab/eval/evaluator.py
+++ b/src/instructlab/eval/evaluator.py
@@ -6,8 +6,8 @@ class Evaluator:
     Parent class for Evaluators
 
     Atttributes:
-        model   The model to be evaluated
+        model_path   Path to the model to be evaluated
     """
 
-    def __init__(self, model: str) -> None:
-        self.model = model
+    def __init__(self, model_path: str) -> None:
+        self.model_path = model_path
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -10,6 +10,9 @@ class EvalError(Exception):
 class ModelNotFoundError(EvalError):
     """
     Exception raised when model is not able to be found
+
+    Attributes
+        model   model that is being operated on
     """
 
     def __init__(self, model) -> None:
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -12,10 +12,13 @@ class ModelNotFoundError(EvalError):
     Exception raised when model is not able to be found
 
     Attributes
-        model   model that is being operated on
+        message     error message to be printed on raise
+        model       model that is being operated on
+        path        filepath of model location
     """
 
-    def __init__(self, model) -> None:
+    def __init__(self, path) -> None:
         super().__init__()
-        self.model = model
-        self.message = f"Model {self.model} could not be found"
+        self.path = path
+        self.model = path.rsplit("/")[-1]
+        self.message = f"Model {self.model} could not be found at {self.path}"
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -1,0 +1,18 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+class EvalError(Exception):
+    """
+    Parent class for all of instructlab-eval exceptions
+    """
+
+
+class ModelNotFoundError(EvalError):
+    """
+    Exception raised when model is not able to be found
+    """
+
+    def __init__(self, model) -> None:
+        super().__init__()
+        self.model = model
+        self.message = f"Model {self.model} could not be found"
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -22,14 +22,10 @@ class MMLU_Evaluator(Evaluator):
         self.few_shots = few_shots
         self.batch_size = batch_size
 
-    def run(self) -> dict:
+    def run(self) -> tuple:
         individual_scores: dict[str, float] = {}
         overall_score: float = 0.0
-        payload = {
-            "individual_scores": individual_scores,
-            "overall_score": overall_score,
-        }
-        return payload
+        return overall_score, individual_scores
 
 
 class PR_MMLU_Evaluator(Evaluator):
@@ -39,8 +35,8 @@ class PR_MMLU_Evaluator(Evaluator):
     Attributes:
         sdg_path    path where all the PR MMLU tasks are stored
         task        group name that is shared by all the PR MMLU tasks
-        few_shots    number of examples
-        batch_size   number of GPUs
+        few_shots   number of examples
+        batch_size  number of GPUs
     """
 
     def __init__(
@@ -57,13 +53,8 @@ class PR_MMLU_Evaluator(Evaluator):
         self.few_shots = few_shots
         self.batch_size = batch_size
 
-    def run(self) -> dict:
+    def run(self) -> tuple:
         individual_scores: dict[str, float] = {}
         overall_score: float = 0.0
         qa_pairs: list[tuple] = []
-        payload = {
-            "individual_scores": individual_scores,
-            "overall_score": overall_score,
-            "qa_pairs": qa_pairs,
-        }
-        return payload
+        return overall_score, individual_scores, qa_pairs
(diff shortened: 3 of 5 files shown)


alinaryan (commented):
This LGTM - are PR-MMLU and PR-Bench going to live in the MMLU and MT-Bench child classes respectively or is the idea to add them in later as their own child classes?

nathan-weinberg:
> This LGTM - are PR-MMLU and PR-Bench going to live in the MMLU and MT-Bench child classes respectively or is the idea to add them in later as their own child classes?

Good question! I think we could go with either approach - either extending the current child classes or adding new child classes but in the same files.

Could you do an explicit approval? I've set the repo so two are required for merge. @alimaredia same for you.

@danmcp and @xukai92 feel free to weigh-in as well - this foundation will define subsequent work.

alinaryan (approved):
I lean towards new child classes since the inputs/dependencies differ across PR benchmarks. You could add them in this PR or it can be handled in a follow-up PR by anyone. 
Thanks for working on this!!

danmcp:
> This LGTM - are PR-MMLU and PR-Bench going to live in the MMLU and MT-Bench child classes respectively or is the idea to add them in later as their own child classes?
> @danmcp and @xukai92 feel free to weigh-in as well - this foundation will define subsequent work.

It would be good to understand the signatures before deciding.  But I'm pretty sure they should be new classes.  The other thing to consider whether the modeling makes more sense as a is-a or has-a relationship.  Has-a might give some useful flexibility.

nathan-weinberg:
> > This LGTM - are PR-MMLU and PR-Bench going to live in the MMLU and MT-Bench child classes respectively or is the idea to add them in later as their own child classes?
> > @danmcp and @xukai92 feel free to weigh-in as well - this foundation will define subsequent work.
> 
> It would be good to understand the signatures before deciding. But I'm pretty sure they should be new classes. The other thing to consider whether the modeling makes more sense as a is-a or has-a relationship. Has-a might give some useful flexibility.

I pushed up two additional child classes based on the specs in the doc - lmk your thoughts!

src/instructlab/eval/evaluator.py
```diff
@@ -0,0 +1,13 @@
...
+
+class Evaluator:
+    """
+    Parent class for Evaluators
+    """
+
+    def __init__(self, model) -> None:
+        self.model = model
+
+    def run(self) -> dict:
```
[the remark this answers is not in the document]
nathan-weinberg: My assumption here was that the generic `run` function would basically take the class attributes and return a single dict for the CLI to consume - thus the following assumption that all Evaluators would return a dict, though with different contents of that dict. 

For example, something like scores and Q&A pairs I was thinking might look like this:
```python
{
    "score": 4.75,
    "qa_pairs": [("question1", "answer1"), ("question2", "answer2")]
}
```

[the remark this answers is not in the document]
danmcp: If the trade off is between sharing a parent method and having more clarity on what's being returned from the function definition I would vote for the latter.

src/instructlab/eval/mtbench.py
```diff
@@ -0,0 +1,32 @@
...
+
+class MT_Bench_Evaluator(Evaluator):
+    """
+    Child class of an Evaluator for Multi-turn Benchmark (MT-Bench)
+
+    Attributes
+        server  vLLM server endpoint
+    """
+
+    def __init__(self, model, server: str) -> None:
```
danmcp, line 15: Should this be server_url

nathan-weinberg: Yeah, I think that's clearer

src/instructlab/eval/evaluator.py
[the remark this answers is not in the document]
nathan-weinberg: Yeah this may not be one that makes sense to inherit

src/instructlab/eval/evaluator.py
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: Apache-2.0
+
+
+class Evaluator:
+    """
+    Parent class for Evaluators
+
+    Atttributes:
+        model   The model to be evaluated
```
danmcp, line 9: Should this be model_path?

nathan-weinberg: Makes sense to me

src/instructlab/eval/mmlu.py
```diff
@@ -0,0 +1,69 @@
...
+from .evaluator import Evaluator
+
+
+class MMLU_Evaluator(Evaluator):
+    """
+    Child class of an Evaluator for Massive Multitask Language Understanding (MMLU)
+
+    Attributes:
+        tasks       list of tasks for MMLU to test the model with
+        fewshots    number of examples
```
danmcp, line 13: few_shots seems to be the most common syntax

src/instructlab/eval/mmlu.py
```diff
@@ -0,0 +1,69 @@
...
+        batchsize   number of GPUs
```
danmcp, line 14: batch_size to be consistent

src/instructlab/eval/mmlu.py
```diff
@@ -0,0 +1,69 @@
...
+        self.tasks = tasks
+        self.fewshots = fewshots
+        self.batchsize = batchsize
+
+    def run(self) -> dict:
+        individual_scores: dict[str, float] = {}
+        overall_score: float = 0.0
+        payload = {
+            "individual_scores": individual_scores,
+            "overall_score": overall_score,
```
danmcp, line 30: Would `return overall_score, individual_scores` be preferable?

nathan-weinberg: I think for @cdoern it would be easier to parse a dict but I'll let him make the final call there

cdoern: I think a single dict makes more sense, less return variability the better

danmcp: Why create a dict just to unpack it?  Plus you can get the type hinting with the tuple.  And wouldn't there be no "parsing" at all with a tuple?

danmcp: > less return variability the better

Is this referring to changes over time not showing up in the definition signature?  The downside though is you are moving your api into the dict keys.  The same breakages will happen on changes, but there is now a naming aspect (of the keys) that needs to be described in the api that won't be evident from the signature.

cdoern: actually, you're right I think I will just be dissecting the dict. The only thing I am wondering is if there's no constant structure, how will I know which values are which?

danmcp: > The only thing I am wondering is if there's no constant structure, how will I know which values are which?

I am not sure I am following.  The call would look something like:

`overall_score, individual_scores = eval.run(...)`

or

`overall_score = eval.run()`

if you only want the first value.  That would be consistent.  If additional return values are ever added to the end, callers would only need to be updated if they want to address the new values.

nathan-weinberg: I pushed up a possible implementation here. Let me both know your thoughts!

src/instructlab/eval/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Exception raised when model is not able to be found
+
+    Attributes
+        model   model that is being operated on
+    """
+
+    def __init__(self, model) -> None:
+        super().__init__()
+        self.model = model
```
danmcp, line 20: model_path

src/instructlab/eval/mmlu.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class MMLU_Evaluator(Evaluator):
+    """
+    Child class of an Evaluator for Massive Multitask Language Understanding (MMLU)
+
+    Attributes:
+        tasks       list of tasks for MMLU to test the model with
+        few_shots    number of examples
+        batch_size   number of GPUs
```
danmcp, line 14: Document return values

src/instructlab/eval/mtbench.py
```diff
@@ -0,0 +1,42 @@
...
+# Local
+from .evaluator import Evaluator
+
+
+class MT_Bench_Evaluator(Evaluator):
+    """
+    Child class of an Evaluator for Multi-turn Benchmark (MT-Bench)
+
+    Attributes
+        server_url  vLLM server endpoint
```
danmcp, line 12: Document return values

src/instructlab/eval/mtbench.py
```diff
@@ -0,0 +1,42 @@
...
+        return overall_score, qa_pairs
+
+
+class PR_Bench_Evaluator(Evaluator):
+    """
+    Child class of an Evaluator for PR-Bench Benchmark (PR-Bench)
+
+    Attributes
+        server_url  vLLM server endpoint
+        questions   questions to be asked
```
danmcp, line 31: Document return values

src/instructlab/eval/exceptions.py
nathan-weinberg: This one I think would remain `model` since the exception refers to a model and not to a path?

danmcp: Is it the model name?  Whatever it was, it couldn't be found.  I was guessing the only thing that would be available to the calling code is that path.

nathan-weinberg: I tweaked the implementation to take the path and dynamically determine the model name for the exception message

src/instructlab/eval/mtbench.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Attributes
+        server_url  vLLM server endpoint
+    """
+
+    def __init__(self, model_path, server_url: str) -> None:
+        super().__init__(model_path)
+        self.server_url = server_url
+
+    def run(self) -> tuple:
```
danmcp, line 19: Ali needed to change this to gen_answers and judge_answers.

nathan-weinberg: Can we merge this and iterate on it? I'd rather just have the code change subsequently then try to align with something currently in-flight.

danmcp: The other option is to include these commits with the impl commits + any needed modifications.

nathan-weinberg: This is just a skeleton PR - not meant to be all of the implementation as well. I think it makes much more sense to merge this now - it also allows @alinaryan to start from the same point to start the MMLU implementation.

danmcp: Good point about blocking mmlu.  I'm good to merge as is.

danmcp (approved):

PR #9, opened 2024-06-17
Add MT-Bench and PR-Bench Support

danmcp:

PR #10, opened 2024-06-19
Implement `MMLU_Evaluator.run()`

alinaryan:
This PR updates the MMLUEvaluator child class to include the lm-evaluation-harness lib dependency and adds an API call to simple_evaluate() to run MMLU. 

ToDo:
- [x] test running in a VM
- [x]  observe the output of results to help decide how to access individual scores vs. overall score
- [x] finish writing the helper function to calculate the overall score based on the results
- [x] parameterize these 57 tasks (I think CLI needs to do this?) https://github.com/instructlab/evaluation/blob/main/scripts/run_mmlu.sh#L24

### DIFF
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -4,7 +4,7 @@
 import os
 
 # Third Party
-from lm_eval.evaluator import simple_evaluate
+from lm_eval.evaluator import simple_evaluate  # type: ignore
 
 # First Party
 from instructlab.eval.evaluator import Evaluator
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -27,6 +27,7 @@ class MMLUEvaluator(Evaluator):
         self.model_dtype = model_dtype
         self.few_shots = few_shots
         self.batch_size = batch_size
+        self.model_path = model_path
 
     def run(self) -> tuple:
         """
@@ -68,6 +69,7 @@ class MMLUBranchEvaluator(Evaluator):
     Child class of an Evaluator for Massive Multitask Language Understanding Branch (MMLUBranch)
 
     Attributes:
+        model_path  absolute path to or name of a huggingface model
         sdg_path    path where all the MMLUBranch tasks are stored
         task        group name that is shared by all the MMLUBranch tasks
         few_shots   number of examples
@@ -82,6 +84,7 @@ class MMLUBranchEvaluator(Evaluator):
         few_shots: int = 2,
         batch_size: int = 5,
     ) -> None:
+        super().__init__()
         self.model_path = model_path
         self.sdg_path = sdg_path
         self.task = task
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -3,6 +3,8 @@
 # Local
 from .evaluator import Evaluator
 
+# Third Party 
+from lm_eval.evaluator import simple_evaluate
 
 class MMLU_Evaluator(Evaluator):
     """
@@ -15,9 +17,11 @@ class MMLU_Evaluator(Evaluator):
     """
 
     def __init__(
-        self, model_path, tasks: list[str], few_shots: int = 2, batch_size: int = 5
+        self, model, model_args, model_path, tasks: list[str], few_shots: int = 2, batch_size: int = 5
     ) -> None:
-        self.model_path = model_path
+        super().__init__(model_path)
+        self.model = model
+        self.model_args = model_args
         self.tasks = tasks
         self.few_shots = few_shots
         self.batch_size = batch_size
@@ -32,8 +36,37 @@ class MMLU_Evaluator(Evaluator):
         """
         individual_scores: dict[str, float] = {}
         overall_score: float = 0.0
+        results = lm_eval.simple_evaluate(
+            model=self.model,
+            model_args=self.model_args,
+            tasks=self.tasks,
+            num_fewshot=self.few_shots,
+            batch_size=self.batch_size,
+            log_samples=True,
+        )
+        #TODO: see what the output of results looks like 
+        #print(results)
+        #calculate_overall_score(results)
         return overall_score, individual_scores
+    
+    def calculate_overall_score(scores):
+        pass # Placeholder for calculating overall score:
+             # overall score = (num model answered correctly / num questions)
 
+############# Testing Code Follows ##############
+def main():
+    # TODO: change this- cli uses HuggingFace to access the model 
+    model = "hf"
+    model_args = "pretrained=$MODEL_PATH,dtype=bfloat16"
+    # Path to the granite model in the aliryan vm on AWS
+    model_path = "/home/ec2-user/instructlab/models/instructlab/granite-7b-lab"
+    #TODO: all 57 tasks need to be parameterized possibly by CLI
+    tasks = "mmlu_abstract_algebra"
+    mmlu = MMLU_Evaluator(model, model_args, model_path, tasks, 2, 5)
+
+if __name__ == "__main__":
+    main()
+############# Testing Code Ends ##############
 
 class PR_MMLU_Evaluator(Evaluator):
     """
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -1,28 +1,30 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Local
-from .evaluator import Evaluator
+from instructlab.eval.evaluator import Evaluator
 
 # Third Party 
 from lm_eval.evaluator import simple_evaluate
+import os
 
 class MMLU_Evaluator(Evaluator):
     """
     Child class of an Evaluator for Massive Multitask Language Understanding (MMLU)
 
     Attributes:
+        model_path   absolute path to or name of a huggingface model
         tasks        list of tasks for MMLU to test the model with
+        model_dtype  dtype of model when served
         few_shots    number of examples
         batch_size   number of GPUs
     """
 
     def __init__(
-        self, model, model_args, model_path, tasks: list[str], few_shots: int = 2, batch_size: int = 5
+        self, model_path, tasks: list[str], model_dtype = 'bfloat16', few_shots: int = 2, batch_size: int = 5
     ) -> None:
         super().__init__(model_path)
-        self.model = model
-        self.model_args = model_args
         self.tasks = tasks
+        self.model_dtype = model_dtype
         self.few_shots = few_shots
         self.batch_size = batch_size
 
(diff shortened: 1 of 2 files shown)


alimaredia:
If anyone needs to test this PR locally, after initializing the virtual env in the repo and installing all the dependencies this python script executes the function:
```
from instructlab.eval.mmlu import MMLU_Evaluator

tasks = ["mmlu_anatomy","mmlu_astronomy"]

# this path is to the granite-7b-lab model on huggingface not on the local filesystem 
model_path = "instructlab/granite-7b-lab"
mmlu = MMLU_Evaluator(model_path, tasks)
overall_score, individual_scores = mmlu.run()
print(overall_score)
print(individual_scores)
```

src/instructlab/eval/mmlu.py
```diff
@@ -30,10 +36,49 @@ def run(self) -> tuple:
...
+    #TODO: all 57 tasks need to be parameterized possibly by CLI
+    tasks = ["mmlu_abstract_algebra","mmlu_anatomy","mmlu_astronomy"]
+    dtype = "float16"
+    mmlu = MMLU_Evaluator(model_path, tasks, dtype, 2, 5)
+    overall_score, individual_scores = mmlu.run()
+    print(overall_score)
+    print(individual_scores)
 
+if __name__ == "__main__":
+    main()
```
JamesKunstle, line 80: You might find it convenient to use `fire.Fire` for some of this in-lib testing code.

e.g.

```
import fire

def main(
    model_path: str = "instructlab/granite-7b-lab", 
    tasks: list[str] = None, 
    dtype: str = "bf16"):
    pass

if __name__ == "__main__":
    fire.Fire(main)
```

then you can invoke the script with

`python3 -m instructlab.train.mmlu --model_path <something>`

fire handles parsing args and passing them to the function / module that you wrap.

src/instructlab/eval/mmlu.py
```diff
@@ -1,24 +1,30 @@
...
         few_shots    number of examples
         batch_size   number of GPUs
     """
 
     def __init__(
-        self, model_path, tasks: list[str], few_shots: int = 2, batch_size: int = 5
+        self, model_path, tasks: list[str], model_dtype = 'bfloat16', few_shots: int = 2, batch_size: int = 5
     ) -> None:
         super().__init__(model_path)
         self.tasks = tasks
```
JamesKunstle, line 26: There should maybe be a default for tasks, overridable by the constructor param. 

In an ideal world, a user might just call:

```python=
MODEL_PATH = args.model_name_or_path # or a model object, might be more convenient for our library

# with defaults for tasks, model type from model object, sensible defaults otherwise
mmlu = MMLU_Evaluator(model_path) 
overall_score, individual_scores = mmlu.run()
```

JamesKunstle: Elsewhere, there could be a `MMLUTasks` object something like:

```
from pydantic import BaseModel, Field

class MMLUTasks(BaseModel):
    mmlu_abstract_algebra: bool = Field(default=True)
    mmlu_anatomy: bool = Field(default=True)
    mmlu_astronomy: bool = Field(default=True)
    ....
```

Then the user could specify which tasks to include vs. not.

src/instructlab/eval/mmlu.py
```diff
@@ -30,10 +36,49 @@ def run(self) -> tuple:
...
+    
+
+############# Testing Code Follows ##############
+def main():
+    # Path to the granite model in the aliryan vm on AWS
+    model_path = "/home/ec2-user/instructlab/models/instructlab/granite-7b-lab"
+    #TODO: all 57 tasks need to be parameterized possibly by CLI
+    tasks = ["mmlu_abstract_algebra","mmlu_anatomy","mmlu_astronomy"]
+    dtype = "float16"
+    mmlu = MMLU_Evaluator(model_path, tasks, dtype, 2, 5)
```
JamesKunstle, line 74: I like this interface a lot

JamesKunstle (changes requested):
very nice PR, makes sense to me. couple of nits.

src/instructlab/eval/mmlu.py
JamesKunstle: oh the evaluation harness handles grabbing the model, nvm I see about the "model_name_or_path thing.

src/instructlab/eval/mmlu.py
```diff
@@ -1,24 +1,30 @@
...
+# Third Party 
+from lm_eval.evaluator import simple_evaluate
+import os
 
 class MMLU_Evaluator(Evaluator):
     """
     Child class of an Evaluator for Massive Multitask Language Understanding (MMLU)
 
     Attributes:
+        model_path   absolute path to or name of a huggingface model
```
danmcp, line 15: I changed model_path to model_name in #9.  If you are going to need model_path, we'll likely need to remove it from the Evaluator parent class.  I guess the question is who will be doing the serving.

src/instructlab/eval/mmlu.py
```diff
@@ -30,10 +36,49 @@ def run(self) -> tuple:
...
+
+############# Testing Code Follows ##############
+def main():
+    # Path to the granite model in the aliryan vm on AWS
+    model_path = "/home/ec2-user/instructlab/models/instructlab/granite-7b-lab"
+    #TODO: all 57 tasks need to be parameterized possibly by CLI
+    tasks = ["mmlu_abstract_algebra","mmlu_anatomy","mmlu_astronomy"]
+    dtype = "float16"
+    mmlu = MMLU_Evaluator(model_path, tasks, dtype, 2, 5)
+    overall_score, individual_scores = mmlu.run()
```
danmcp, line 75: I changed run to more specific methods in #9.  We don't have to here but it could leave room for other functions in the future.  Using run seems to imply the object really only does 1 thing.

src/instructlab/eval/mmlu.py
nathan-weinberg: @alinaryan unless there's a reason for divergence I think we should stay consistent if possible and use `model_name`

src/instructlab/eval/mmlu.py
```diff
@@ -30,10 +36,49 @@ def run(self) -> tuple:
             overall_score       MMLU score for the overall model evaluation
             individual_scores   Individual MMLU score for each task
         """
-        individual_scores: dict[str, float] = {}
-        overall_score: float = 0.0
+        #TODO: make this a parameter for class?
+        os.environ['TOKENIZERS_PARALLELISM'] = 'true'
```
nathan-weinberg, line 40: I think this could be a CLI config, either via flag or profile, @cdoern wdyt?

src/instructlab/eval/mmlu.py
```diff
@@ -30,10 +36,49 @@ def run(self) -> tuple:
...
+            agg_score += float(mmlu_res['acc,none'])
+            individual_scores[task] = {}
+            individual_scores[task]['score'] = float(mmlu_res['acc,none'])
+            individual_scores[task]['stderr'] = float(mmlu_res['acc_stderr,none'])
+
+        overall_score = float(agg_score/len(self.tasks))
         return overall_score, individual_scores
+    
+
+############# Testing Code Follows ##############
```
nathan-weinberg, line 67: Could we add this to a separate file like @danmcp did in #9?

That way we can run it via the work that will be done in #12

src/instructlab/eval/mmlu.py
alimaredia: `model_path` here is used differently than in MT-Bench. `model_path` is can be the path on the filesystem to the locally downloaded model. This model can be downloaded via `ilab download --repository=instructlab/granite-7b-lab`.

The `model_path` can also be the name of the repo and the model on huggingface. For example just passing ing `instructlab/granite-7b-lab` would download the model from https://huggingface.co/instructlab/granite-7b-lab.

If we want as a follow up maybe we ensure that the `model_path` is only for a model downloaded locally, but I don't think that should hold up this PR.

Hope this clarity helps on why the name is as is.

src/instructlab/eval/mmlu.py
alimaredia: @JamesKunstle we'd discussed this quite a bit when we first started working on the library and decided we want the caller to the library to pass in the tasks list and not have a default tasks list or something preset.

src/instructlab/eval/mmlu.py
danmcp: My point was that if we do continue to need a full model path here, we should just remove it from Evaluator since it's not common between mtbench and mmlu.

src/instructlab/eval/mmlu.py
alimaredia: @xukai92 if you know could you explain what TOKENIZERS_PARALLELISM is giving us? The reason I added this is because the output when I ran without setting that environment var was giving me warning to set that environment variable. 

For the sake of moving this PR along I say we just make a low priority issue for this TODO and someone can investigate this further.

src/instructlab/eval/mmlu.py
alimaredia: We should absolutely move this into a separate file! That being said I think we should do it in a follow up PR. We should just remove this `Testing Code Follows` section from the PR.

alimaredia:
I've mentioned this in some of my comments in code review, Getting mmlu working would be a big deal for the library so I propose we do the following and then merge the PR ASAP:

1. Remove the testing bits from `mmlu.py`
2. Someone besides the authors (Alina or I) test the PR by hand, to see the scores being output properly
3. Create follow up issues for anything else that we know needs to be done.

src/instructlab/eval/mmlu.py
cdoern: this should be in the eval profile, but for now will be a cli flag.

click has a way to read from env vars into flags too

src/instructlab/eval/mmlu.py
alimaredia: Yeah it looks like we'll need to remove it from the Evaluator base class. Should we do that in this PR?

src/instructlab/eval/mmlu.py
alinaryan: Thank you - I'm going to skip this for now since it has already been tested. I appreciate the suggestion

src/instructlab/eval/mmlu.py
alinaryan: @cdoern is going to handle parameterizing the defaults in the cli. Also, a user won't need to worry about changing the defaults - it should always be a standardized set for MMLU, but I could see that case being made for MMLUBench

src/instructlab/eval/mmlu.py
danmcp: That works.  I'll make the corresponding adjustment in #9.

alinaryan: I agree about staying consistent, but it does need to be model_path bc that's what the CLI expects.

src/instructlab/eval/mmlu.py
alinaryan: That makes sense, I am going to keep it to run for now and could change this in a follow-up PR.

src/instructlab/eval/mmlu.py
alinaryan: I didn't see the above discussion until after I posted my comment, but this sounds good! I'll remove it from the base class too

nathan-weinberg: Sounds good y'all!

src/instructlab/eval/mmlu.py
nathan-weinberg: That's fine with me! Will give me a chance to setup runners anyway

alinaryan: Removed the testing code! Will re-add into a separate file in a followup PR like Ali suggested.

src/instructlab/eval/mmlu.py
```diff
@@ -1,24 +1,30 @@
...
 
     def __init__(
-        self, model_path, tasks: list[str], few_shots: int = 2, batch_size: int = 5
+        self, model_path, tasks: list[str], model_dtype = 'bfloat16', few_shots: int = 2, batch_size: int = 5
     ) -> None:
         super().__init__(model_path)
         self.tasks = tasks
+        self.model_dtype = model_dtype
         self.few_shots = few_shots
         self.batch_size = batch_size
```
cdoern, line 30:
```suggestion
self.model_path = model_path
```
```

**Review by cdoern — COMMENTED** [review]
missing model path, run tox -e ruff as well! looks good though, I will approve after those changes

**Comment by nathan-weinberg** [comment]
@alinaryan can you rebase?

File: src/instructlab/eval/mmlu.py (hunk 7 above)
Comment (alinaryan) [reply src/instructlab/eval/mmlu.py]: Added issue here: https://github.com/instructlab/eval/issues/15

File: src/instructlab/eval/mmlu.py [hunk 9]
```diff
@@ -30,18 +36,40 @@ def run(self) -> tuple:
...
+        overall_score = float(agg_score/len(self.tasks))
+        return overall_score, individual_scores
 
-class PR_MMLU_Evaluator(Evaluator):
+class MMLUBranchEvaluator(Evaluator):
     """
-    Child class of an Evaluator for PR Massive Multitask Language Understanding (PR MMLU)
+    Child class of an Evaluator for Massive Multitask Language Understanding Branch (MMLUBranch)
 
     Attributes:
```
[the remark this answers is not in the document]
alinaryan: Added!

src/instructlab/eval/mmlu.py
```diff
@@ -30,18 +45,42 @@ def run(self) -> tuple:
...
             individual_scores   Individual MMLU score for each task
         """
-        individual_scores: dict[str, float] = {}
-        overall_score: float = 0.0
+        # TODO: make this a parameter for class?
+        os.environ["TOKENIZERS_PARALLELISM"] = "true"
+
+        individual_scores: dict = {}
+        agg_score: float = 0.0
+        model_args = "pretrained=" + self.model_path + ",dtype=" + self.model_dtype
```
danmcp, line 53: Nit: Use f""

src/instructlab/eval/mmlu.py
alinaryan: Added follow-up issue here: https://github.com/instructlab/eval/issues/16

alimaredia (approved):

JamesKunstle (approved):

danmcp (approved):

PR #14, opened 2024-06-25
Add e2e job to CI

### LINKED ISSUE #13: Add E2E test workflow to CI
Similar to https://github.com/instructlab/sdg/pull/33

nathan-weinberg:
Resolves #13

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,10 +31,17 @@ jobs:
 
     steps:
       # No step-security/harden-runner since this is a self-hosted runner
-      - name: Checkout
+      - name: Checkout instructlab/eval
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
           # https://github.com/actions/checkout/issues/249
+          fetch-depth: 0
+
+      - name: Checkout instructlab/instructlab
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: "instructlab/instructlab"
+          path: "instructlab"
           fetch-depth: 0
 
       - name: Install Packages
@@ -68,7 +75,6 @@ jobs:
           export PATH="/home/runner/.local/bin:/usr/local/cuda/bin:$PATH"
           python3 -m venv venv
           . venv/bin/activate
-          git clone https://github.com/instructlab/instructlab.git
           cd instructlab
           sed 's/\[.*\]//' requirements.txt > constraints.txt
           python3 -m pip cache remove llama_cpp_python
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,92 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E test
+
+on:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+          sudo apt-get install -y cuda-toolkit git cmake build-essential virtualenv
+          nvidia-smi
+          ls -l /dev/nvidia*
+
+      - name: Setup Python 3.11
+        uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+        with:
+          python-version: 3.11
+          cache: pip
+          cache-dependency-path: |
```
bjhargrave, line 51: This will only include the deps of this repo and not instructlab which is cloned later. You want to checkout the cli before this step (as above).

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+    permissions:
+      pull-requests: write
+
+    steps:
+      # No step-security/harden-runner since this is a self-hosted runner
+      - name: Checkout
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          # https://github.com/actions/checkout/issues/249
+          fetch-depth: 0
```
bjhargrave, line 38:
```suggestion
      - name: Checkout
        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
        with:
          # https://github.com/actions/checkout/issues/249
          fetch-depth: 0
      - name: Checkout instructlab/instructlab
        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
        with:
          repository: "instructlab/instructlab"
          path: "instructlab"
          fetch-depth: 0
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+          path: ~/.cache/huggingface
+          # config contains DEFAULT_MODEL
+          key: huggingface-${{ hashFiles('src/instructlab/configuration.py') }}
+
+      - name: Install instructlab and instructlab-eval
+        run: |
+          export PATH="/home/runner/.local/bin:/usr/local/cuda/bin:$PATH"
+          python3 -m venv venv
+          . venv/bin/activate
+          git clone https://github.com/instructlab/instructlab.git
```
bjhargrave, line 71:
```suggestion
```

bjhargrave (changes requested):

nathan-weinberg:
@bjhargrave do you think the suggested changes here should be done for the SDG implementation as well?

bjhargrave:
> do you think the suggested changes here should be done for the SDG implementation as well?

Yes. You want all the checkouts done before pip caching to get all the cache-key files in place.

I detest using caching in CI jobs since it is a source of innumerable errors. But I can see that uncached installs of the cli take a while. So we should at least try to get the caching right.

danmcp (approved):

alinaryan (approved):

danmcp:
I think the e2e tests might need this change: https://github.com/instructlab/eval/pull/17/files

PR #17, opened 2024-06-25
Updating to > 1.0 openai

danmcp:
Updating to a newer openai to not conflict with the cli

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 FastChat
 GitPython>=3.1.42,<4.0.0
 shortuuid
-openai<1.0.0
+openai>=1.13.3,<2.0.0
 psutil
 torch
 transformers
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -50,8 +50,7 @@ class MMLUEvaluator(Evaluator):
 
         individual_scores: dict = {}
         agg_score: float = 0.0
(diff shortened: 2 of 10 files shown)


src/instructlab/eval/mmlu.py
```diff
@@ -50,8 +50,7 @@ def run(self) -> tuple:
 
         individual_scores: dict = {}
         agg_score: float = 0.0
-        model_args = f"pretrained= {self.model_path}, dtype= {self.model_dtype}"
-
+        model_args = f"pretrained={self.model_path},dtype={self.model_dtype}"
```
danmcp, line 53: @alimaredia @alinaryan FYI in case you hit this issue as well.

nathan-weinberg (approved):

requirements.txt
```diff
@@ -2,7 +2,7 @@
 FastChat
 GitPython>=3.1.42,<4.0.0
 shortuuid
-openai<1.0.0
```
JamesKunstle, line 5: are there any implications to this bump? or is it a clean reversioning?

JamesKunstle (approved):
everything changes here looks good if it works during manual testing. one question about versioning implications but not something that should block merging.

requirements.txt
danmcp: I might not be understanding the questions.  But there were several changes with the API that needed corresponding changes on our side that are in the commit.

alinaryan (approved):

PR #19, opened 2024-06-26
mmlu branch run() complete

alimaredia:


src/instructlab/eval/mmlu.py
```diff
@@ -104,11 +107,36 @@ def run(self) -> tuple:
...
-            qa_pairs            Question and answer pairs from the evaluation
+            overall_score       Average MMLUBranch score for the task group
+            individual_scores   Individual MMLUBranch scores for each task in the task group
         """
-        individual_scores: dict[str, float] = {}
-        overall_score: float = 0.0
-        qa_pairs: list[tuple] = []
-        return overall_score, individual_scores, qa_pairs
+        #TODO: make this a parameter for class?
+        os.environ['TOKENIZERS_PARALLELISM'] = 'true'
```
danmcp, line 114: Most of this code looks the same as the non branch version.  Can it be moved to a common function?

src/instructlab/eval/mmlu.py
```diff
@@ -104,11 +107,36 @@ def run(self) -> tuple:
...
+            model="hf",
+            model_args=model_args,
+            tasks=self.tasks,
+            num_fewshot=self.few_shots,
+            batch_size=self.batch_size,
+            task_manager=tm,
+        )
+
+        results = mmlu_output["results"]
+        print(results)
```
danmcp, line 132: We probably want to leave the display of the caller.

danmcp (changes requested):

src/instructlab/eval/mmlu.py
```diff
@@ -104,11 +107,36 @@ def run(self) -> tuple:
...
-        individual_scores: dict[str, float] = {}
-        overall_score: float = 0.0
-        qa_pairs: list[tuple] = []
-        return overall_score, individual_scores, qa_pairs
+        #TODO: make this a parameter for class?
+        os.environ['TOKENIZERS_PARALLELISM'] = 'true'
+
+        individual_scores: dict = {}
+        agg_score: float = 0.0
+        model_args = "pretrained=" + self.model_path + ",dtype=" + self.model_dtype
```
nathan-weinberg, line 118: Can we use an f-string here?

danmcp: This is part of the code that should be common with the non branch version. Which was switched to use fstring already.

src/instructlab/eval/mmlu.py
nathan-weinberg: Will address in a followup

JamesKunstle (approved):
lgtm

alinaryan (approved):
This LGTM. We should run the old version on a particular checkpoint, record the number, and then run the new version to make sure everything is the same, like Kai suggested. If it differs, it can be fixed in a followup PR, but let's get this in for now.

danmcp (approved):
Approving but noting the future cleanup.

nathan-weinberg:
> Approving but noting the future cleanup.

I'll open an issue for tracking

PR #21, opened 2024-06-26
add cpu support for MMLU bench

cdoern:
simple_evaluate has a device option that seems to default to cuda. People with 128GB or even ~90GB+ of ram should be able to run eval on CPU. Add auto detection for if torch is available


nathan-weinberg (approved):
`torch` is already in our `requirements.txt` so this should be a fine addition

alinaryan (approved):

PR #34, opened 2024-06-28
e2e: Fix permissions error

russellb:
The e2e jobs started failing today on a permissions error with this
command. The output was useful when debugging at one point, so I left
it in place, but now run it with sudo to avoid the permissions error.

Signed-off-by: Russell Bryant <russell.bryant@gmail.com>


danmcp (approved):
LGTM, Thanks!

PR #40, opened 2024-06-28
Use a portion of the hexadecimal string

### LINKED ISSUE #38: question_id is not be stored consistently and is probably losing consistency
Currently the question_id is generated as a string here:

https://github.com/instructlab/eval/blob/main/src/instructlab/eval/mt_bench_branch_generator.py#L71

Because it is too large for an int32.  But when it is returned through qa_pairs, question_ids come out looking like: 7.969730277787438e+37

Not sure yet whether it's losing the type in the dataframes or json parsing in 

https://github.com/instructlab/eval/blob/main/src/instructlab/eval/mt_bench_judgment.py

One fix would be to find a smaller consistent hash.  Otherwise need to figure out the typing to keep it a str throughout.

alinaryan:
Takes a subset of the hex string to produce a smaller integer.

Resolves #38

### DIFF
--- a/src/instructlab/eval/mt_bench_branch_generator.py
+++ b/src/instructlab/eval/mt_bench_branch_generator.py
@@ -68,7 +68,7 @@ def generate(judge_model_name, branch, taxonomy_dir, output_dir):
                 # Generate a consistent hash to have consistent question_id across qna_files from different runs
                 str_bytes = bytes(q, "UTF-8")
                 m = hashlib.md5(str_bytes)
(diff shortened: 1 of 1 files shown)


danmcp:
I put together this fix: https://github.com/instructlab/eval/pull/42. There might still be a better hashing algorithm we could use that will allow us to use an int throughout.

nathan-weinberg:
Looking at both I'd probably favor #42 to keep the hashing precision, is that alright with you @alinaryan?

alinaryan:
@nathan-weinberg Yes

PR #41, opened 2024-06-30
Allow first_n option for gen answers, fix return values with max_workers=1 and only print api errors on final failure

danmcp:
Three minor changes/fixes:

- INSTRUCTLAB_EVAL_FIRST_N_QUESTIONS is now supported for gen answers in addition to judgment.  This is to enable faster sniff test cycles
- Fixing an error where no result was returned for max_workers == 1.  Return was at the wrong indent level.
- Only print the failing error from api calls.

### DIFF
--- a/src/instructlab/eval/mt_bench_answers.py
+++ b/src/instructlab/eval/mt_bench_answers.py
@@ -116,9 +116,17 @@ def generate_answers(
     if os.path.isfile(answer_file):
         os.remove(answer_file)
 
+    first_n = None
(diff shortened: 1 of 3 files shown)


nathan-weinberg (approved):

alimaredia (approved):
After running `export INSTRUCTLAB_EVAL_FIRST_N_QUESTIONS=10`
I see gen_answers only running 10 questions: 

```
INFO:     127.0.0.1:51902 - "POST /v1/chat/completions HTTP/1.1" 200 OK
 20%|████████████████████████████████████████▍                                                                                                                                                                 | 2/10 [00:19<01:12,  9.09s/it]
 ```

PR #46, opened 2024-07-02
Remove task categories from mmlu tasklist

alinaryan:
Having task categories listed inside the mmlu task list was breaking the mmlu benchmark given task categories duplicate the individual tasks. This commit removes task categories from the list to keep it just to individual tasks.

Discussion started here: https://redhat-external.slack.com/archives/C077AMX9MSR/p1719897501712589


danmcp (approved):

nathan-weinberg (approved):

JamesKunstle:
This is exactly what I was hoping for. It'd be awesome if we could extend the mmlu test to accommodate the entire MMLU list as well to check for little glitches like this over the entire task-list.

PR #48, opened 2024-07-03
Add additional exceptions and error handling

### LINKED ISSUE #36: Error handling
The library currently doesn't have robust error handling. We should implement this with our custom error classes and do appropriate handling on the CLI side.

nathan-weinberg:
Resolves #36

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -22,3 +22,77 @@ class ModelNotFoundError(EvalError):
         self.path = path
         self.model = path.rsplit("/")[-1]
         self.message = f"Model {self.model} could not be found at {self.path}"
+
+
+class MMLUEvaluationError(EvalError):
+    """
+    Exception raised when MMLU evaluation of a model is not able to be completed
+
+    Attributes
+        message     error message to be printed on raise
+        cause       root cause exception
+        model       model being evaluated
+    """
+
+    def __init__(self, model, cause) -> None:
+        super().__init__()
+        self.model = model
+        self.cause = cause
+        self.message = f"Failed to run MMLU evaluation on {self.model}: {self.cause}"
+
+
+class MMLUBranchEvaluationError(EvalError):
+    """
+    Exception raised when MMLUBranch evaluation of a model is not able to be completed
+
+    Attributes
+        message     error message to be printed on raise
+        cause       root cause exception
+        model       model being evaluated
+        sdg_path    filepath of SDG data
+    """
+
+    def __init__(self, model, sdg_path, cause) -> None:
+        super().__init__()
+        self.model = model
+        self.sdg_path = sdg_path
+        self.cause = cause
+        self.message = f"Failed to run MMLUBranch evaluation on {self.model} with {self.sdg_path}: {self.cause}"
+
+
+class MTBenchEvaluationError(EvalError):
+    """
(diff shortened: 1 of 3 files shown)


danmcp:
Generally looks good so far.  My question is how far we want to go?  For example, what happens when you pass invalid model, judge, branch, base_branch, taxonomy, ...?

danmcp:
Note: Several of the unit tests here: https://github.com/instructlab/instructlab/pull/1572

Are starting to hit some of the ugly error handling.

nathan-weinberg:
> Generally looks good so far. My question is how far we want to go? For example, what happens when you pass invalid model, judge, branch, base_branch, taxonomy, ...?

Thanks! And that's a good question - I wasn't sure how deep to go with this for a first pass. I think there's some errors that will live in the lib itself and some that make more sense to do on the CLI side. @alinaryan @alimaredia @cdoern y'all have any thoughts around this?

danmcp:
> I wasn't sure how deep to go with this for a first pass. I think there's some errors that will live in the lib itself and some that make more sense to do on the CLI side

Many/most of the issues will require logic on both sides.  The examples from the unit tests (bad branch or bad taxonomy dir) will both need a better error sent from the backend before the cli can catch those exceptions and output a nice message.

danmcp:
After working on some more specific error cases in
https://github.com/instructlab/instructlab/pull/1616 and https://github.com/instructlab/eval/pull/52

I am thinking this PR isn't going down the right path because:
- Catching general exceptions in mtbench and mmlu and returning a fairly general eval error will mask and specific eval errors that are thrown up to that level
- I don't think the top level exception like mtbenchevalerror helps over just having EvalError.  Having more specific errors Ex: GitRepoNotFound is probably all we need to extend from EvalError.  I wouldn't for example want a git repo not found error specific to mtbench and mtbenchbranch.
- We probably shouldn't catch any unknown exceptions at the level of this library.  I think our goal should be to cover all the known exceptions with nice EvalErrors and messages.  And let the caller handle those and general exceptions as they see fit.  In the cli for example, I would suggest we handle all EvalErrors with an appropriate message to the user.  But let other exception fall through.  With a goal of finding all the exception cases and handling appropriately.

Note: Passing cause on the constructor rather than with from will give pylint errors

danmcp (changes requested):
After working on some more specific error cases in
https://github.com/instructlab/instructlab/pull/1616 and https://github.com/instructlab/eval/pull/52

I am thinking this PR isn't going down the right path because:

Catching general exceptions in mtbench and mmlu and returning a fairly general eval error will mask and specific eval errors that are thrown up to that level
I don't think the top level exception like mtbenchevalerror helps over just having EvalError. Having more specific errors Ex: GitRepoNotFound is probably all we need to extend from EvalError. I wouldn't for example want a git repo not found error specific to mtbench and mtbenchbranch.
We probably shouldn't catch any unknown exceptions at the level of this library. I think our goal should be to cover all the known exceptions with nice EvalErrors and messages. And let the caller handle those and general exceptions as they see fit. In the cli for example, I would suggest we handle all EvalErrors with an appropriate message to the user. But let other exception fall through. With a goal of finding all the exception cases and handling appropriately.
Note: Passing cause on the constructor rather than with from will give pylint errors

nathan-weinberg:
@danmcp and I chatted offline and the code has progressed while I was on PTO to a point it makes more sense to close this than retool it

PR #59, opened 2024-07-09
[mt-bench] Add missing return val

alinaryan:
`error_rate` was missing from the return list and erroring out

### DIFF
--- a/tests/test_judge_answers.py
+++ b/tests/test_judge_answers.py
@@ -5,7 +5,7 @@ import pprint
 from instructlab.eval.mt_bench import MTBenchEvaluator
 
 mt_bench = MTBenchEvaluator("instructlab/granite-7b-lab", "instructlab/granite-7b-lab")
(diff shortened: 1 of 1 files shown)


tests/test_judge_answers.py
```diff
@@ -5,7 +5,7 @@
 from instructlab.eval.mt_bench import MTBenchEvaluator
 
 mt_bench = MTBenchEvaluator("instructlab/granite-7b-lab", "instructlab/granite-7b-lab")
-overall_score, qa_pairs, turn_scores = mt_bench.judge_answers(
+overall_score, qa_pairs, turn_scores, error_rate = mt_bench.judge_answers(
```
danmcp, line 8: Would be good to print the error rate

danmcp (changes requested):
Need the same change for test_branch_judge_answers

danmcp:
Resolved with https://github.com/instructlab/eval/pull/63

Issue #60, opened 2024-07-09
Invalid sdg path with mmlu_branch prints stack trace and unobvious error


danmcp:
If you run:

ilab model evaluate --model models/instructlab/granite-7b-lab --benchmark mmlu_branch --sdg-path invalid --base-model models/instructlab/granite-7b-lab 

You get:

Traceback (most recent call last):
  File "/home/ec2-user/instructlab/venv/bin/ilab", line 8, in <module>
    sys.exit(ilab())
             ^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ec2-user/instructlab/src/instructlab/utils.py", line 551, in wrapper
    return func(*args, **kwargs)
[...]

PR #61, opened 2024-07-09
Error handling with sdg_path not found and invalid

danmcp:
This PR is adding error handling for sdg_path not found.  To do so, I have unified the simple_evaluate calls between mmlu and mmlu_branch into an "abstract" parent class.

Resolves: #60 

Will add unit tests in the CLI after this PR merges

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -9,7 +9,7 @@ class EvalError(Exception):
 
 class ModelNotFoundError(EvalError):
     """
-    Exception raised when model is not able to be found
+    Error raised when model is not able to be found
 
     Attributes
         message     error message to be printed on raise
@@ -18,12 +18,13 @@ class ModelNotFoundError(EvalError):
 
     def __init__(self, path) -> None:
         super().__init__()
+        self.path = path
         self.message = f"Model could not be found at {path}"
 
 
 class InvalidGitRepoError(EvalError):
     """
-    Exception raised when taxonomy dir provided isn't a valid git repo
+    Error raised when taxonomy dir provided isn't a valid git repo
     Attributes
         message         error message to be printed on raise
         taxonomy_dir    supplied taxonomy directory
@@ -31,12 +32,13 @@ class InvalidGitRepoError(EvalError):
 
     def __init__(self, taxonomy_dir) -> None:
         super().__init__()
+        self.taxonomy_dir = taxonomy_dir
         self.message = f"Invalid git repo: {taxonomy_dir}"
 
 
 class GitRepoNotFoundError(EvalError):
     """
-    Exception raised when taxonomy dir provided does not exist
+    Error raised when taxonomy dir provided does not exist
     Attributes
         message         error message to be printed on raise
         taxonomy_dir    supplied taxonomy directory
@@ -44,12 +46,13 @@ class GitRepoNotFoundError(EvalError):
 
     def __init__(self, taxonomy_dir) -> None:
         super().__init__()
+        self.taxonomy_dir = taxonomy_dir
         self.message = f"Taxonomy git repo not found: {taxonomy_dir}"
 
 
 class InvalidGitBranchError(EvalError):
     """
-    Exception raised when branch provided is invalid
+    Error raised when branch provided is invalid
     Attributes
         message         error message to be printed on raise
         branch          supplied branch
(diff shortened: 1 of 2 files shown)


danmcp:
I've added a test case here: https://github.com/instructlab/instructlab/compare/main...danmcp:sdgpatherrors?expand=1. Will wait till this merges before submitting the PR there to avoid the failing test in the interim.

src/instructlab/eval/mmlu.py
```diff
@@ -206,20 +210,29 @@ def run(self) -> tuple:
             individual_scores   Individual MMLUBranch scores for each task in the task group
         """
         logger.debug(locals())
+
+        if not os.path.exists(self.sdg_path):
+            raise SDGPathNotFoundError(self.sdg_path)
```
markmc, line 215: Hard to get this stuff perfect, but worth seeing if you're ok with what happens if the file isn't readable (in terms of permissions)

src/instructlab/eval/exceptions.py
```diff
@@ -58,3 +58,29 @@ class InvalidGitBranchError(EvalError):
...
         super().__init__()
         self.message = f"Invalid git branch: {branch}"
+
+
+class SDGPathNotFoundError(EvalError):
+    """
+    Error raised when the sdg path doesn't exist
+    Attributes
+        message         error message to be printed on raise
+        sdg_path        sdg path
```
markmc, line 68: I think it would be good to make it an actual attribute, not just a constructor argument

src/instructlab/eval/mmlu.py
```diff
@@ -206,20 +210,29 @@ def run(self) -> tuple:
...
+            results = run_mmlu(
+                self.model_path,
+                self.model_dtype,
+                self.tasks,
+                self.few_shots,
+                self.batch_size,
+                self.sdg_path,
+            )
+        except KeyError as ke:
+            if len(self.tasks) > 0 and f"'{self.tasks[0]}'" in str(ke):
```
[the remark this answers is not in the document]
markmc: Ah, I can answer that myself from the #60 

```
File "/home/ec2-user/eval/src/instructlab/eval/mmlu.py", line 89, in run_mmlu
mmlu_output = simple_evaluate(
^^^^^^^^^^^^^^^^
...
File "/home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/lm_eval/tasks/init.py", line 133, in _get_tasklist
return self.task_index[name]["task"]
~~~~~~~~~~~~~~~^^^^^^
KeyError: 'mmlu_pr'
```

I'd be inclined to do something like:

```
# If first_task_key file cannot be found in sdg_path, simple_evaluate() will return
# an obscure KeyError(first_task_key) - this function converts that specific error
# into a more user-understandable error
def _simple_evaluate_with_error_handling(tasks_dir, first_task_key, **kwargs):
    try:    
        return simple_evaluate(**kwargs)
    except KeyError as ke:
        if ke.args[0] == first_task_key:
            raise InvalidSDGPathError(tasks_dir) from ke
        raise
```

src/instructlab/eval/mmlu.py
danmcp: For the directory itself, it's already checked above: Error: Invalid value for '--sdg-path': Path 'tests/testdata/mmlu_branch' is not readable.

But I'll go ahead and add a check here since the library is generic.

src/instructlab/eval/mmlu.py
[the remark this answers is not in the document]
danmcp: I think I got the essence of this suggestion in my latest version.  Let me know if not.

src/instructlab/eval/mmlu.py
danmcp: Fixed, thanks for the suggestions!

src/instructlab/eval/exceptions.py
danmcp: Fixed, thanks!

src/instructlab/eval/mmlu.py
```diff
@@ -167,36 +207,21 @@ def run(self) -> tuple:
...
-        batch_size  number of GPUs
+        model_path      absolute path to or name of a huggingface model
+        sdg_path        path where the <TASK_NAME>.jsonl and <TASK_NAME>_task.yaml files for the branches being evaluated are stored
+        tasks           group name that is shared by all the MMLUBranch tasks
+        model_dtype     dtype of model when served
+        few_shots       number of examples
+        batch_size      number of GPUs
     """
 
     name = "mmlu_branch"
```
JamesKunstle, line 223: what's this for? I don't see it referenced elsewhere

JamesKunstle (approved):
one comment but PR seems like it can go in once tests pass.

src/instructlab/eval/mmlu.py
danmcp: https://github.com/instructlab/eval/commit/f6ac6566c2710a0f06bb390447d4c4968c5dec5a

markmc (approved):
Goodness, exception handling can be quite the rabbit hole!

alinaryan (approved):

PR #62, opened 2024-07-11
[mmlu] Allow optionally setting a PyTorch device

alinaryan:
Gives the ability to optionally specify a PyTorch device when running MMLU and MMLUBranch, defaults to cpu.

### DIFF
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -79,7 +79,7 @@ MMLU_TASKS = [
 
 
 def run_mmlu(
-    model_path, model_dtype, tasks, few_shots, batch_size, sdg_path=None
+    model_path, model_dtype, tasks, few_shots, batch_size, device, sdg_path=None
 ) -> dict:
     try:
         model_args = f"pretrained={model_path},dtype={model_dtype}"
@@ -92,7 +92,7 @@ def run_mmlu(
             tasks=tasks,
             num_fewshot=few_shots,
             batch_size=batch_size,
-            device=("cuda" if torch.cuda.is_available() else "cpu"),
+            device=device,
             task_manager=tm,
         )
         results = mmlu_output["results"]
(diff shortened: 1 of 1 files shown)


cdoern (approved):
looks good. lets add this to the cli cfg, and as a flag to eval

src/instructlab/eval/mmlu.py
```diff
@@ -79,7 +79,7 @@
 
 
 def run_mmlu(
-    model_path, model_dtype, tasks, few_shots, batch_size, sdg_path=None
+    model_path, model_dtype, tasks, few_shots, batch_size, device, sdg_path=None
```
danmcp, line 82: Can we make device optional?  Otherwise it's an api breaking change.

danmcp (changes requested):

src/instructlab/eval/mmlu.py
alinaryan: I set a default in the cli https://github.com/instructlab/instructlab/pull/1672 but can change if we think it should be optional

alinaryan: Recap from eval meeting today: I will change this PR to make 'device' optional and I will leave https://github.com/instructlab/instructlab/pull/1672 open with a hold for now

src/instructlab/eval/mmlu.py
```diff
@@ -114,13 +114,15 @@ class MMLUEvaluator(Evaluator):
...
         batch_size   number of GPUs
+        device       PyTorch device (e.g. "cpu" or "cuda:0") for running models
     """
 
     name = "mmlu"
 
     def __init__(
         self,
         model_path,
+        device: str = ("cuda" if torch.cuda.is_available() else "cpu"),
```
cdoern, line 125: this makes sense to me

cdoern (commented):
looks good. So basically its a option one can pass to the class and the defaulting that previously existed has been moved up to _init_

missing an `import torch`

danmcp (approved):

PR #64, opened 2024-07-11
Rename sdg_path to tasks_dir

danmcp:
Related to https://github.com/instructlab/instructlab/issues/1663

Corresponding UI change: https://github.com/instructlab/instructlab/pull/1677

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -64,29 +64,29 @@ class InvalidGitBranchError(EvalError):
         self.message = f"Invalid git branch: {branch}"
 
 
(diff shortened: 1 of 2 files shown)


nathan-weinberg (approved):

markmc (approved):
I guess eval isn't tested in the e2e CI tests yet, so we don't need to worry about doing this in multiple steps to avoid breaking CI?

danmcp:
> I guess eval isn't tested in the e2e CI tests yet, so we don't need to worry about doing this in multiple steps to avoid breaking CI?

That's correct, the only API change was to the exceptions which aren't used directly.

jaideepr97 (approved):
LGTM!

Issue #12, opened 2024-06-24
Add test runner to CI


nathan-weinberg:
@danmcp has introduced some tests in #9 - we should get testing setup as part of our current CI stack once they are merged in

nathan-weinberg:
Note for this we will need a GPU-enabled runner with the following specs:
- 24GB vram for MT-Bench
- 96GB vram for MMLU

russellb:
> Note for this we will need a GPU-enabled runner with the following specs:
> 
> * 24GB vram for MT-Bench

This job should be capable:

https://github.com/instructlab/instructlab/pull/1556

I can help get the same thing running on this repo, but it will take some additional repo configuration first.

> * 96GB vram for MMLU

do you have an AWS instance type in mind?  Ideally the minimal / cheapest type capable of testing this

danmcp:
For the simple flow test, 24GB of vram should work for both.  I haven't tried to make it work with less.  We've been using g5.4xlarge.

alimaredia:
I think we can make CI work with a `g4dn.2xlarge` instance. I'm going to give it a try today and report back on this issue. If not the g5.4xlarge should definitely work.

booxter:
@alimaredia any updates on this issue? It would be nice to validate patches in the repo through CI before merging.

alimaredia:
@booxter I don't have any updates. I've been focused on the instructlab CI lately. If someone wants to take the lead on establishing CI within the eval repo over the next week or so I'd be happy to review or tell that person what I see the next steps are.

PR #71, opened 2024-07-16
[WIP] Add new GPU runner for E2E job, incorporate unit tests into existing runner

nathan-weinberg:
This PR does the following:
- Adds a new GPU runner job (AWS-based) we can use for full E2E runs (current runner was just sanity checking integration)
- Changes existing GPU runner to run our unit test suite
- Some modifications to unit tests so they can be run in CI

Resolves https://github.com/instructlab/eval/issues/12

### DIFF
--- a/.github/workflows/e2e-nvidia-a10g-x1.yml
+++ b/.github/workflows/e2e-nvidia-a10g-x1.yml
@@ -1,0 +1,205 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: E2E (NVIDIA A10G x1)
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_or_branch:
+        description: 'pull request number or branch name'
+        required: true
+        default: 'main'
+
+jobs:
+  start-runner:
+    name: Start external EC2 runner
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@fcfb31a5760dad1314a64a0e172b78ec6fc8a17e # v2.3.6
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ami-00c51d9c1374eda97
+          ec2-instance-type: g5.2xlarge
+          subnet-id: subnet-02d230cffd9385bd4
+          security-group-id: sg-06300447c4a5fbef3
+          iam-role-name: instructlab-ci-runner
+          aws-resource-tags: >
+            [
+              {"Key": "Name", "Value": "instructlab-ci-github-runner"},
+              {"Key": "GitHubRepository", "Value": "${{ github.repository }}"}
+            ]
+
+  e2e:
+    name: E2E Test
+    needs: start-runner
+    runs-on: ${{ needs.start-runner.outputs.label }}
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Checkout instructlab/eval
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          # https://github.com/actions/checkout/issues/249
+          fetch-depth: 0
+
+      - name: Checkout instructlab/instructlab
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          repository: "instructlab/instructlab"
+          path: "instructlab"
+          fetch-depth: 0
+
+      - name: Determine if pr_or_branch is a PR number
+        id: check_pr
+        run: |
+          nvidia-smi
+          if [[ "${{ github.event.inputs.pr_or_branch }}" =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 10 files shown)


.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -0,0 +1,197 @@
...
+        if: steps.check_pr.outputs.is_pr == 'false'
+        run: |
+          git checkout ${{ github.event.inputs.pr_or_branch }}
+
+      - name: Install Packages
+        run: |
+          cat /etc/os-release
+          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
+
+      - name: Install ilab
```
russellb, line 119: You need to add a `git clone` of instructlab in this section for running this in the eval repo

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -0,0 +1,197 @@
...
+
+          python3.11 -m pip cache purge
+
+          DS_BUILD_CPU_ADAM=1 BUILD_UTILS=1 python3.11 -m pip install deepspeed
+
+          nvidia-smi
+
+          python3.11 -m pip show nvidia-nccl-cu12
+          
+          ./scripts/basic-workflow-tests.sh -cemf
```
russellb, line 161:
```suggestion
          cd instructlab
          ./scripts/basic-workflow-tests.sh -cemf
```

nathan-weinberg: Going to remove `c` arg as well since it doesn't seem to do anything

russellb: yeah that was an old option

README.md
```diff
@@ -1,6 +1,7 @@
 # eval
 
 ![Lint](https://github.com/instructlab/eval/actions/workflows/lint.yml/badge.svg?branch=main)
+![Test](https://github.com/instructlab/eval/actions/workflows/test.yml/badge.svg?branch=main)
```
alimaredia, line 4: @nathan-weinberg on line 38, 39 of this file can you mention how to run the tests with pytest individually and as a group?

nathan-weinberg: Yes!

.github/workflows/test.yml
```diff
@@ -67,34 +72,35 @@ jobs:
...
-          python3 -m pip install bitsandbytes
+          python3.11 -m pip install bitsandbytes
           # install instructlab
-          python3 -m pip install .
-          cd ..
-          # Install instructlab-eval
-          python3 -m pip install .
+          python3.11 -m pip install .
+          # start llama-cpp server
+          ilab model download --repository instructlab/granite-7b-lab-GGUF --filename granite-7b-lab-Q4_K_M.gguf
```
alimaredia, line 89: Since what we want to do with `ilab` is pretty basic, I'd suggest we just install instructlab from pypi.

.github/workflows/test.yml
```diff
@@ -67,34 +72,35 @@ jobs:
...
+          ilab model serve --model-path /home/runner/.local/share/instructlab/models/granite-7b-lab-Q4_K_M.gguf
```
alimaredia, line 90: In the functional tests scripts we shut down the server to clean up. I wonder if we should split all of this into a shell script that manages ilab installation, server startup, pytest running, and server shutdown.

nathan-weinberg: Really, we shouldn't need `ilab` at all to run unit or functional tests for this library - it should operate independently of the CLI

.github/workflows/test.yml
nathan-weinberg: I would agree except the current pypi package is pretty out-of-date - using it now is just going to require us to change a bunch of stuff when the next release comes out - but once `0.18.0` lands that'll make sense

.github/workflows/test.yml
nathan-weinberg: I'm gonna look into some other possible approaches

.github/workflows/test.yml
alinaryan: I agree we should install from pypi, installing instructlab from pypi won't change the ilab model download command

.github/workflows/test.yml
alinaryan: > Really, we shouldn't need ilab at all to run unit or functional tests for this library - it should operate independently of the CLI

I think this makes sense for unit tests, but functional ones should prob still use a server from ilab etc

alinaryan (commented):
Good stuff

.github/workflows/test.yml
jaideepr97: @nathan-weinberg we can install the beta releases with `--pre` and then remove that once 0.18.0 lands

.gitignore
```diff
@@ -53,6 +53,7 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 cover/
+eval_output/
```
booxter, line 56: `taxonomy` too?

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -0,0 +1,205 @@
...
+
+      - name: Install Packages
+        run: |
+          cat /etc/os-release
+          sudo dnf install -y gcc gcc-c++ make git python3.11 python3.11-devel
+
+      - name: Install ilab
+        run: |
+          export CUDA_HOME="/usr/local/cuda"
+          export LD_LIBRRY_PATH="$LD_LIBRARY_PATH:/usr/local/cuda/lib64:/usr/local/cuda/extras/CUPTI/lib64"
```
booxter, line 129: `LIBRRY` :)

.github/workflows/e2e-nvidia-a10g-x1.yml
```diff
@@ -0,0 +1,205 @@
...
+
+          # TODO This should be added to instructlab-training
+          python3.11 -m pip install packaging wheel
+
+          python3.11 -m pip install instructlab-training[cuda]
+
+          # Install the local version of eval before installing the CLI so PR changes are included
+          python3.11 -m pip install .
+
+          python3.11 -m pip install instructlab
```
booxter, line 147: (Question) AFAIU this assumes that `instructlab` package never caps / pins the `eval` library in an incompatible way. Otherwise, this line could revert the `eval` package to the one from `pypi`. Is it an acceptable assumption?

nathan-weinberg:
closing in favor of https://github.com/instructlab/eval/pull/155

Issue #77, opened 2024-07-22
`test_branch_gen_answers` does not fail when no model is being served


khaledsulayman:
In trying to work on the eval CI I noticed that the library doesn't raise any errors when there is no model being hosted at the requested port.

As I understand it, our current OpenAI error handling prints exceptions to stdout instead of failing because of some expected behavior with the API where it may be ok to fail temporarily while we keep retrying. However, currently, we are catching `openai.OpenAIError`, which is more general than `open.APIConnectionError`, which is what we are actually seeing in this case (no model being served).

I believe we can fix this one of two ways:
1. keep the general except clause and if the error is `open.APIConnectionError`, we raise it. We could do this either immediately or after the max retries are completed.
2. if applicable, we could just specify the error we're seeing that is requiring us to catch it and retry a few times. If I remember correctly this is a rate limiting issue so I'd imagine there is a separate native `openai` exception type for this scenario but would need to do more digging to confirm if that's the only scenario in which we'd want to carry out this retry functionality.

khaledsulayman:
Here are the OpenAI library error types: https://help.openai.com/en/articles/6897213-openai-library-error-types-guidance

It seems like we may be able to specifically use `APIError`, `Timeout`, `RateLimitError`, etc. to match the specific cases where we'd want to wait and retry, but might be worth further discussing which of these are relevant in this case.

PR #80, opened 2024-07-23
Don't retry on connection error

booxter:
Note: TimeoutError is a subclass of the generic connection error, and we'd like to retry for timeouts.

This patch also rearranges the code a bit, incl. making sure that it won't sleep at the very last iteration when we know there won't be another retry attempt.

Closes: https://github.com/instructlab/eval/issues/77

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ scripts/infra/cloud-instance.sh ec2 ssh
 # Regardless of how you setup your instance
 git clone https://github.com/instructlab/taxonomy.git && pushd taxonomy && git branch rc && popd
 git clone --bare https://github.com/instructlab/eval.git && git clone eval.git/ && cd eval && git remote add syncrepo ../eval.git
-python -m venv venv
+python3 -m venv venv
 source venv/bin/activate
 pip install -r requirements.txt
 pip install -r requirements-dev.txt
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -90,3 +90,15 @@ class InvalidTasksDirError(EvalError):
         super().__init__()
         self.tasks_dir = tasks_dir
         self.message = f"Invalid Tasks Dir: {tasks_dir}"
+
+
+class OpenAIError(EvalError):
+    """
+    Error raised when reply retrieval from OpenAI API fails.
+    Attributes
+        message              error message to be printed on raise
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.message = "Failed to receive a reply from API."
--- a/src/instructlab/eval/mt_bench_common.py
+++ b/src/instructlab/eval/mt_bench_common.py
@@ -17,6 +17,9 @@ import time
 from fastchat.model.model_adapter import get_conversation_template  # type: ignore
 import openai
 
+# First Party
+from instructlab.eval import exceptions
+
 # Local
 from .logger_config import setup_logger
 
@@ -247,10 +250,15 @@ def play_a_match_single(
     return result
 
 
+def _is_fatal_openai_error(e: openai.OpenAIError) -> bool:
+    return isinstance(e, openai.APIConnectionError)
+
+
 def chat_completion_openai(
     openai_client, model, conv, temperature, max_tokens, merge_system_user_message=False
 ) -> str:
-    output = API_ERROR_OUTPUT
+    output = None
+
     for i in range(API_MAX_RETRY):
         try:
             messages = conv.to_openai_api_messages()
@@ -272,14 +280,38 @@ def chat_completion_openai(
             )
             output = response.choices[0].message.content
             break
-        except openai.OpenAIError as e:
+        except (
+            # retry may help with these errors
+            openai.APIConnectionError,
+            openai.RateLimitError,  # 429
+            openai.InternalServerError,  # >=500
+            # NOTE: Errors listed below may need a revisit: we are not sure if
+            # it's ever helpful to retry them. Leaving them intact for now.
+            openai.AuthenticationError,  # 401
+            openai.PermissionDeniedError,  # 403
+            openai.NotFoundError,  # 404
+        ) as e:
+            if not _is_fatal_openai_error(e):
+                output = API_ERROR_OUTPUT  # disable hard fail (never raise!)
+            # still, retry in the hope we'll get a successful reply
             if i == API_MAX_RETRY - 1:
                 # Print error on last try
                 print(type(e), e)
             else:
                 logger.debug(e)
             time.sleep(API_RETRY_SLEEP)
+        except (
+            # retry won't fix these errors
+            openai.BadRequestError,  # 400
+            openai.UnprocessableEntityError,  # 422
+        ) as e:
+            logger.debug(e)
+            return API_ERROR_OUTPUT  # immediately soft fail
 
+    if output is None:
+        # not a single attempt was non-fatal; this is indicative of
+        # basic connectivity or server issue -> hard fail
+        raise exceptions.OpenAIError
     return output
 
 
--- a/src/instructlab/eval/mt_bench_common.py
+++ b/src/instructlab/eval/mt_bench_common.py
@@ -4,7 +4,7 @@ Common data structures and utilities.
 """
 
 # Standard
-from typing import Optional
+from typing import Optional, TypedDict
 import ast
 import dataclasses
 import glob
@@ -14,6 +14,7 @@ import re
 import time
 
 # Third Party
+from fastchat import conversation
 from fastchat.model.model_adapter import get_conversation_template  # type: ignore
 import openai
 
@@ -254,23 +255,44 @@ def _is_fatal_openai_error(e: openai.OpenAIError) -> bool:
     return isinstance(e, openai.APIConnectionError)
 
 
+# TODO: copied from instructlab (cli) utils module; consolidate somewhere?
+class Message(TypedDict):
+    """
+    Represents a message within an AI conversation.
+    """
+
+    content: str
+    # one of: "user", "assistant", or "system"
+    role: str
+
+
+def _get_messages(
+    conv: conversation.Conversation, merge_system_user_message: bool
+) -> list[Message]:
+    messages = conv.to_openai_api_messages()
+    if (
+        merge_system_user_message
+        and messages[0]["role"] == "system"
+        and messages[1]["role"] == "user"
+    ):
+        messages[1]["content"] = messages[0]["content"] + "\n" + messages[1]["content"]
+        return messages[1:]
+    return messages
+
+
 def chat_completion_openai(
-    openai_client, model, conv, temperature, max_tokens, merge_system_user_message=False
+    openai_client,
+    model,
+    conv: conversation.Conversation,
+    temperature,
+    max_tokens,
+    merge_system_user_message: bool = False,
 ) -> str:
     output = None
+    messages = _get_messages(conv, merge_system_user_message)
 
     for i in range(API_MAX_RETRY):
         try:
-            messages = conv.to_openai_api_messages()
-            if (
-                merge_system_user_message
-                and messages[0]["role"] == "system"
-                and messages[1]["role"] == "user"
-            ):
-                messages[1]["content"] = (
-                    messages[0]["content"] + "\n" + messages[1]["content"]
-                )
-                messages = messages[1:]
             response = openai_client.chat.completions.create(
                 model=model,
                 messages=messages,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,3 +96,6 @@ from-first = true
 # import-heading-firstparty=First Party
 # import-heading-localfolder=Local
 known-local-folder = ["tuning"]
+
+[tool.mypy]
+ignore_missing_imports = true


README.md
```diff
@@ -21,7 +21,7 @@ scripts/infra/cloud-instance.sh ec2 ssh
 # Regardless of how you setup your instance
 git clone https://github.com/instructlab/taxonomy.git && pushd taxonomy && git branch rc && popd
 git clone --bare https://github.com/instructlab/eval.git && git clone eval.git/ && cd eval && git remote add syncrepo ../eval.git
-python -m venv venv
+python3 -m venv venv
```
booxter, line 24: `python` outside a virtualenv doesn't always exist, or point to `python3`. Better to be explicit.

booxter:
@khaledsulayman

README.md
khaledsulayman: seems out of scope of this PR, although it's not a big one so I'm not opposed to including it if you don't think it's an issue

src/instructlab/eval/mt_bench_common.py
```diff
@@ -270,17 +273,16 @@ def chat_completion_openai(
...
-        except openai.OpenAIError as e:
-            if i == API_MAX_RETRY - 1:
-                # Print error on last try
-                print(type(e), e)
-            else:
-                logger.debug(e)
-            time.sleep(API_RETRY_SLEEP)
-
-    return output
+            return response.choices[0].message.content
```
khaledsulayman, line 276: won't this return on the first iteration? Please correct me if I'm missing something

README.md
booxter: Sorry, yes. It sneaked in. I can drop it or split out into a separate commit.

src/instructlab/eval/mt_bench_common.py
booxter: Yes. This is what the code does right now, no? It `break`s our of the `for` loop, which makes it immediately `return output`. I think that's by design - the function is banging at the API endpoint until a successful reply comes back, then returns the reply. Only in the unfortunate cases of exceptions we - sometimes - want to retry again.

alimaredia:
I think one of the reasons for being liberal with the exception handling at this level is that it's not uncommon in runs of MT-Bench to see 400's get returned randomly in a sea of 200's.

Screenshot from 2024-07-25 09-47-54


That being said it seems like the errors you're handling are not 400s so this fix seems helpful.
 @danmcp any thoughts here?

src/instructlab/eval/mt_bench_common.py
```diff
@@ -270,17 +273,16 @@ def chat_completion_openai(
...
+        except openai.APITimeoutError as ex:
+            logger.debug(ex)
+        except openai.APIConnectionError as ex:
```
danmcp, line 279: One error condition we should be able to create is overloading llama-cpp cpu serving with max-workers.  That tends to create a service unavailable condition.  Would be good to know where that hits in here.

Also, when calling against the openai api, do we know for sure this isn't a condition we hit during normal use?  They were retrying up to 16 times previously but I don't know the common errors returned.

src/instructlab/eval/mt_bench_common.py
```diff
@@ -250,9 +252,10 @@ def play_a_match_single(
 def chat_completion_openai(
     openai_client, model, conv, temperature, max_tokens, merge_system_user_message=False
 ) -> str:
-    output = API_ERROR_OUTPUT
```
danmcp, line 253: It's unclear whether it's appropriate to remove this condition vs allow it to be included in the error rate.  While I prefer this strict interpretation, it would potentially create a lot of flakiness in the results.  So it's a tradeoff of completeness/accuracy vs flakiness.

src/instructlab/eval/mt_bench_common.py
booxter: I honestly don't know how often 40x return from API. I will rely on your judgement. (Which I think suggests it's a common scenario.)

---

My primary goal is to make the code that calls this function to fail (to report back to CI that it's broken). Without it, e2e runs claim success when in reality evaluation crashed. I think CI status should be indicative of a failure - one should not have to drill down into logs to double-check a job actually worked.

But: I think this can be achieved while leaving the retry logic for `APIConnectionError` in place. Just need to `raise` instead of `return output`. This will allow us to retry for some time but still fail if an error happened.

What do you think?

khaledsulayman:
@danmcp what if we were to add some handling on the final retry and raise something more descriptive of what actually went wrong? Whether we want to ignore a one-off error during a single run is one thing, but if we hit max retries and especially if it's all the same error, I think it makes sense to have enough granularity to be able to easily see what went wrong.

danmcp:
> @danmcp what if we were to add some handling on the final retry and raise something more descriptive of what actually went wrong? Whether we want to ignore a one-off error during a single run is one thing, but if we hit max retries and especially if it's all the same error, I think it makes sense to have enough granularity to be able to easily see what went wrong.

As an example of an issue that's easy to hit is the context length of your judge model is too small.  Something around 5120 is required by mt-bench.  In those cases, for the conversations that require larger than your judge model supports, it's going to fail consistently on those questions.  We then have a choice of failing and raising the exception or failing and counting in the error rate.  The original logic ignored the error and so we added the error rate to at least account for it.  And other errors are possible too such as the model might not give an answer in the expected format.  If we change the logic to fail fast on errors, it could throw away hours of runtime.  If everything is perfectly tuned, this might be fine.  But we would need to prove that can be accomplished reliably before we can make such a change.

I think some easy wins are:
- Catch specific exceptions as is being done in this PR which indicate the system isn't working at all.  We just need to make sure the exceptions indicate complete failure and not flakiness.
- Adding an option to not raise exceptions on a particular class of problems.  Ex: context length of the model isn't long enough.  Anything that implied you weren't going to get valid results could fit into this bucket.  We need something like this for testing purposes to be able to use smaller models with testing.
- Add tuning for the number of retries to attempt before giving up.

booxter:
@danmcp "Adding an option to not raise exceptions on a particular class of problems. Ex: context length of the model isn't long enough." I think the issue is that there's no specific type to catch for such particular error. (But we can check what we have.)

I think in general - 1) the library should raise on error; 2) the caller should decide what to do with the error (log, crash, etc.) We can make callers track exceptions and, at the end of the eval, use the fact that any errors happened as the reason to exit with non-zero value (which should tell the test script it failed.)

btw I noticed that openai library apparently already retries and there's a tunable to control how frequently it does. https://github.com/openai/openai-python/blob/195c05a64d39c87b2dfdf1eca2d339597f1fce03/src/openai/__init__.py#L117

src/instructlab/eval/mt_bench_common.py
danmcp: I am a +1 to failing fast for any conditions which are known to be invalid.  No model being served is one of those.  As long as we can differentiate that case, let's do it.  The two conditions we know of that we need to make sure still work are:

- Getting 400s because of max context length issues.  You would get this using granite as the judge model for example.
- llama-ccp cpu serving being overloaded and returning service unavailable.

Neither of those conditions should fail fast and we just need to make sure they don't hit this code block.  I would assume the 400 error does not.  But I don't know about the service unavailable.

For the context length issues additional improvements might be:
- Don't retry when you hit the issue the first time but record the error and move on
- Default to failing fast with this condition and allow for an environment variable to override

The case I am not comfortable with yet is failing after the retries run out vs recording the error in the error rate.  There is more discussion on this in response to other comments.

danmcp (changes requested):

danmcp:
> I think in general - 1) the library should raise on error; 2) the caller should decide what to do with the error (log, crash, etc.) We can make callers track exceptions and, at the end of the eval, use the fact that any errors happened as the reason to exit with non-zero value (which should tell the test script it failed.)

I am not quite sure how this would work.  To be able to make the decision on whether to continue on error, we couldn't raise an exception all the way outside the library.  At that point it's too late to continue.  We could have some sort of callback mechanism I guess.

At a high level I think we can all agree it's not great the code is eating errors.  More granular error handling should be able to differentiate the different types of errors and stop eating any errors that indicate something is broken vs hitting iffy configurations vs flakiness in various serving options.  But it's not clear that passing the errors on regardless of type solves the general problem.  The purpose of the library is to execute a benchmark.  In this case, it's a long running benchmark and some of the questions being asked by the benchmark may fail to get a valid response.  The library currently pushes the responsibility to the user to decide whether the error rate is acceptable.

So unless we get to the point where all api errors can be eliminated for long runs, it's not clear yet how we can do much better than isolating the failures which should fail fast (like model not found).

Letting the caller make a decision with a callback, or with exception handling (not sure how it would work still), or with env variable options, or by passing in options to declare what types of error should fail are all options we could support.  I am in favor of using such options to be able to fail for things like the max context length issue by default with an option to override.

If there is a suggestion to fail for any exception after retries, we just need to explain how it's going to work for long running jobs with potentially small error rates as well as work for testing with smaller context length models.

booxter:
For reference, here is the list of `openai` exceptions:

https://github.com/openai/openai-python/blob/195c05a64d39c87b2dfdf1eca2d339597f1fce03/README.md?plain=1#L367-L376

booxter:
Thanks for this, I think I start understanding larger context here. A strict exception discipline won't do when you'd like work to continue even after a bad answer is returned.

---

I think ultimately what's missing is `gen_answers` interface of an evaluator doesn't have any way to communicate the presence of errors in the output data. We could convert the type of the entrypoint to `-> bool`, then use this value to inform the caller that the produced data set has issues.

---

Another venue to explore could be: the library uses file system as interface to pass generated data to the caller. It could instead return the produced dataset (via `yield`?) and let the caller decide if it wants to accept a particular answer. I don't think it's the job of the library to write data onto disk...

---

Ignoring what can be done in this library, the caller could probably open the written file and check if it contains any error markers. If so, make sure an error return code is returned. (When it's ready.)

---

I will take some time to think on the issue. But if you have some preference or thoughts on the above alternatives, please let me know.

danmcp:
> I think ultimately what's missing is `gen_answers` interface of an evaluator doesn't have any way to communicate the presence of errors in the output data. We could convert the type of the entrypoint to `-> bool`, then use this value to inform the caller that the produced data set has issues.

That's a really good point.  I've been thinking about this in terms of the openai api and the error rates we expose from judgment.  But we don't expose error rates on gen answers today.  Giving an error rate might be an appropriate return value?  That would match what we do for judgment.


> 
> Another venue to explore could be: the library uses file system as interface to pass generated data to the caller. It could instead return the produced dataset (via `yield`?) and let the caller decide if it wants to accept a particular answer. I don't think it's the job of the library to write data onto disk...
> 
> Ignoring what can be done in this library, the caller could probably open the written file and check if it contains any error markers. If so, make sure an error return code is returned. (When it's ready.)
> 
> I will take some time to think on the issue. But if you have some preference or thoughts on the above alternatives, please let me know.

As of now, my preference is to:

- Add error rate (or equivalent) to gen_answers for completeness
- Isolate errors from https://github.com/openai/openai-python/blob/195c05a64d39c87b2dfdf1eca2d339597f1fce03/README.md?plain=1#L367-L376 which should fail fast or fail after some number of retries.
Note I checked and we are getting <class 'openai.InternalServerError'> Service Unavailable with llama-cpp and cpu serving.  So I think your current impl of raising on APIConnectionError is appropriate.  I am slightly worried about not having the retries even that case though.  I assume we would get a similar error on network blip.
  - 400 - discussed below.  Either need to allow or an option to allow.
  - 401-404 - fail after retries
  - 422-500 - include in error rate
- Leave the -1 and "$ERROR$" in place for conditions we can't clearly classify as complete failures
- Include your sleep fix

Can be fixed later, if we do we should open an issue
- Create a way for some error codes to be allowed for testing purposes.  Namely allow the 400s.  And switch these cases to fail fast by default.

booxter:
For now, a more limited approach just making the tests fail on errors in the output file: https://github.com/instructlab/eval/pull/85 while we think through a way to pass errors or error rate directly.

booxter:
Also splitting out the sleep fix here: https://github.com/instructlab/eval/pull/84 since it's not directly related to this issue.

khaledsulayman (approved):

src/instructlab/eval/mt_bench_common.py
```diff
@@ -273,13 +294,20 @@ def chat_completion_openai(
             output = response.choices[0].message.content
             break
         except openai.OpenAIError as e:
+            if _is_fatal_openai_error(e):
```
danmcp, line 297: I might not be following this correctly yet.  The logic I was expecting was:

400 - Include in error rate after 1 occurrence
422-500 - Include in error rate after retries
401-404 + APIConnectionError - fail after retries

If we are only trying to handle APIConnectionError, I am not quite following why APIConnectionErrors are being allowed to continue but timeout errors and all other errors seem to be marked as non fatal but will exit below when their output is none.

src/instructlab/eval/mt_bench_common.py
```diff
@@ -273,13 +294,20 @@ def chat_completion_openai(
...
                 print(type(e), e)
             else:
                 logger.debug(e)
             time.sleep(API_RETRY_SLEEP)
 
+    if output is None:
+        # Hard fail only when none of the attempts resulted in a proper
+        # response or a "soft" error (in which case we register the error
+        # in the output file).
+        raise exceptions.CompletionError(messages)
```
danmcp, line 310: This exception feels very specific for the types of errors we could be getting.  Seems useful for a 400 for example.  But probably not for a ConnectionError

danmcp (changes requested):

src/instructlab/eval/mt_bench_common.py
```diff
@@ -273,13 +294,20 @@ def chat_completion_openai(
             output = response.choices[0].message.content
             break
         except openai.OpenAIError as e:
+            if _is_fatal_openai_error(e):
+                output = API_ERROR_OUTPUT
             if i == API_MAX_RETRY - 1:
                 # Print error on last try
                 print(type(e), e)
```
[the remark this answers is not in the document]
booxter: let's handle this separately. https://github.com/instructlab/eval/pull/86

src/instructlab/eval/mt_bench_common.py
booxter: I believe you suggested to "include in error rate" for 400. If so, we won't need to raise anything at all. But yes, the class name should probably indicate it's only about basic connectivity / communication problems and such.

src/instructlab/eval/mt_bench_common.py
danmcp: > But I think **422** is indicative of the input not being accepted. MDN [suggests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) that repeating it won't produce a different result. Should we handle it like 400 then? (meaning: Include in error rate.)

+1, good correction

> 
> > 401-404 - fail after retries.
> 
> What's the point of retrying `AuthenticationError` or `PermissionDeniedError` or `NotFoundError` (especially the last two). Don't we expect that if access is denied, then it will be denied if we retry again? (Same for 404.)

I am iffy on these between fail fast and fail after retries.  My fear was there could be flakiness in these responses.  In which case the only harm that's done with retries is it would wait for 20ish seconds before it declared it a failure and gave up.

booxter: OK I will include these in retry list, with a note that we may explore later if they justify retries.

src/instructlab/eval/mt_bench_common.py
danmcp: Sorry, I picked a bad example.  You are correct a 400 shouldn't make it here.  My concern was the name and the messages content.  I don't know what anyone would do with the messages for the error types that would hit this.

booxter: Yep. I am yanking the `messages` argument completely, this is of no use now that this is raised only for basic connectivity issues.

booxter:
I'm leaving this as draft still until I have a chance to run it against a real server. (Will do it tomorrow.) Code wise, I hope this at least reflects the latest discussions and intent though. (I also included a minor change where I split out `messages` calculation out of the `for` loop; this is a separate commit to simplify review.)

src/instructlab/eval/exceptions.py
```diff
@@ -90,3 +90,15 @@ def __init__(self, tasks_dir) -> None:
         super().__init__()
         self.tasks_dir = tasks_dir
         self.message = f"Invalid Tasks Dir: {tasks_dir}"
+
+
+class OpenAIError(EvalError):
```
danmcp, line 95: Since the underlying API could change, maybe this should be more generic. ModelServingAPIError?

src/instructlab/eval/mt_bench_common.py
```diff
@@ -272,14 +302,38 @@ def chat_completion_openai(
...
             break
-        except openai.OpenAIError as e:
+        except (
+            # retry may help with these errors
+            openai.APIConnectionError,
+            openai.RateLimitError,  # 429
+            openai.InternalServerError,  # >=500
+            # NOTE: Errors listed below may need a revisit: we are not sure if
+            # it's ever helpful to retry them. Leaving them intact for now.
+            openai.AuthenticationError,  # 401
```
danmcp, line 312: I thought we were saying the 40xs would fit into the is fatal category.

src/instructlab/eval/mt_bench_common.py
```diff
@@ -272,14 +302,38 @@ def chat_completion_openai(
             )
             output = response.choices[0].message.content
             break
-        except openai.OpenAIError as e:
```
danmcp, line 275: Don't we still need to handle this exception as a catchall.  In case they add a new exception type for example.

nathan-weinberg:
Closing in favor of https://github.com/instructlab/eval/pull/103

PR #84, opened 2024-07-29
Don't sleep on last retry attempt

booxter:
Completion attempt was retrying a number of times if API returned errors, sleeping for some time in between attempts. Due to a coding mistake, it was doing a final "sleep" after the final attempt; then returned error regardless. This sleep was redundant. The PR removes it.


danmcp (approved):

khaledsulayman (approved):

nathan-weinberg:
@booxter could you add a PR description just for future reference?

PR #85, opened 2024-07-29
tests: Fail when model is not served

booxter:
Parse the answers file and exit with non-zero return code when files contain an error flag choice.

This is a more limited fix (just for tests) while we are figuring out a better design to pass errors and error rates to the caller.

Closes: #77

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,0 +1,21 @@
+# First Party
+from instructlab.eval.mt_bench_common import (
+    API_ERROR_OUTPUT,
+    bench_dir,
+    load_model_answers,
+)
+
+
+def _is_bad_answer(a):
+    return any(API_ERROR_OUTPUT in choice["turns"] for choice in a["choices"])
+
+
+def is_bad_answer_file(evaluator):
+    branch = getattr(evaluator, 'branch', None)
+    base_dir = bench_dir(evaluator.output_dir, evaluator.name, branch)
+    answer_dir = f"{base_dir}/model_answer"
+    for dataset in load_model_answers(answer_dir).values():
+        for answer in dataset.values():
+            if _is_bad_answer(answer):
+                return True
+    return False
--- a/tests/test_branch_gen_answers.py
+++ b/tests/test_branch_gen_answers.py
@@ -1,3 +1,9 @@
+# Standard
+import sys
+
+# Third Party
+from common import is_bad_answer_file
+
 # First Party
 from instructlab.eval.mt_bench import MTBenchBranchEvaluator
 
@@ -8,3 +14,6 @@ mt_bench_branch = MTBenchBranchEvaluator(
     "main",
 )
 mt_bench_branch.gen_answers("http://localhost:8000/v1")
+
+if is_bad_answer_file(mt_bench_branch):
+    sys.exit(1)
--- a/tests/test_gen_answers.py
+++ b/tests/test_gen_answers.py
@@ -1,5 +1,14 @@
+# Standard
+import sys
+
+# Third Party
+from common import is_bad_answer_file
+
 # First Party
 from instructlab.eval.mt_bench import MTBenchEvaluator
 
 mt_bench = MTBenchEvaluator("instructlab/granite-7b-lab", "instructlab/granite-7b-lab")
 mt_bench.gen_answers("http://localhost:8000/v1")
+
+if is_bad_answer_file(mt_bench):
+    sys.exit(1)


tests/common.py
```diff
@@ -0,0 +1,21 @@
...
+    load_model_answers,
+)
+
+
+def _is_bad_answer(a):
+    return any(API_ERROR_OUTPUT in choice["turns"] for choice in a["choices"])
+
+
+def is_bad_answer_file(evaluator):
+    branch = getattr(evaluator, 'branch', None)
```
danmcp, line 14: Use ""

tests/common.py
```diff
@@ -0,0 +1,21 @@
+# First Party
+from instructlab.eval.mt_bench_common import (
+    API_ERROR_OUTPUT,
+    bench_dir,
+    load_model_answers,
+)
+
+
+def _is_bad_answer(a):
+    return any(API_ERROR_OUTPUT in choice["turns"] for choice in a["choices"])
```
danmcp, line 10: Won't we always hit this condition using granite or other insufficient judge models

tests/test_branch_gen_answers.py
```diff
@@ -8,3 +14,6 @@
     "main",
 )
 mt_bench_branch.gen_answers("http://localhost:8000/v1")
+
+if is_bad_answer_file(mt_bench_branch):
```
danmcp, line 18: Maybe use assert?

danmcp (changes requested):

tests/common.py
booxter: I am clueless. :) @danmcp do you have something to read up on why judge models would consistently produce bad answers?

danmcp: I don't have a document. There are two different cases of producing bad answers.

1) Some models will return api errors for the longer conversations.  In particular models that don't have max context lengths >= 5120 won't be able to complete all the evaluation conversations.

2) Some models that haven't been trained as judges will not consistently return a rating in the requested format.  Granite and merlinite fit into this category of being inconsistent.

Both of these mean if we want to be able to run through the flow with judges that don't fully work for testing purposes, errors are expected.

booxter: This comes down to - what are these test cases even supposed to validate: the presence of errors? the ability to connect to models at all? if we only want the latter, there should be a way to distinguish between an "error in reply" and "connection issue". Since we don't want to raise an exception (at least right now) for an error, maybe we could use different error messages to distinguish?.. So you'd have not just API_ERROR_OUTPUT but also API_ERROR_CONNECTION etc.

danmcp: I am not sure they are really test cases in their current state.  They were initially added as a way to exercise the library.  Since they were added, pretty much all of the focus has been in the instructlab cli to perform the validation of the e2e.  (No objection to turning them into something more useful)

From: https://github.com/instructlab/eval/pull/80 I was thinking we were headed to a point where the fatal errors would fail outright (after some number of retries).  Seems like a good change to better classify whatever error conditions are left.

booxter: Thanks for clarifications.

There's https://github.com/instructlab/eval/issues/12 that asks to enable the tests in CI. Maybe I should fix this first, esp. since now I realize that maybe my change breaks the test scripts with granite (I do not have access to hardware and was not able to run them with `vllm` server running. I should.)

---

We can still explore #80 and, additionally, a new API that would allow to expose errors to callers; I was not sure if your `I am slightly worried about not having the retries even that case though. I assume we would get a similar error on network blip.` means that you still want to continue retrying on `APIConnectionError` (which is not what #80 does.)

If we **expect** errors in the answers file, then this PR here is invalid.

danmcp: > We can still explore #80 and, additionally, a new API that would allow to expose errors to callers; I was not sure if your `I am slightly worried about not having the retries even that case though. I assume we would get a similar error on network blip.` means that you still want to continue retrying on `APIConnectionError` (which is not what #80 does.)

Correct, I was suggesting we split the errors into two buckets:

- Fatal errors we retry and fail after max attempts
- Errors which we record after max attempts


> 
> If we **expect** errors in the answers file, then this PR here is invalid.

Apologies, I didn't notice you were only checking for answers after gen answers and not after judgement.  So the code as is would probably work with granite/merlinite.  Errors in answers are much less common than errors in judgement.  But we are currently in a state of needing to be tolerant of some errors.  If it's ok, let's resolve what errors should be possible to exist here before we validate non exist.

danmcp: Also, if we do make the change to just return the error rate from gen_answers as we do with judgment.  The assert could just be to check that the error_rate is 0 rather than use the file parsing.

danmcp:
The root issue was addressed with https://github.com/instructlab/eval/pull/103 and this test will now fail with an exception if a model is not being served.  No objection we want to improve the tests further.  Closing now due to inactivity and other related fixes.

PR #86, opened 2024-07-31
Remove remaining print()s from the library

booxter:
Use logger for any messages produced instead.

### DIFF
--- a/src/instructlab/eval/mt_bench_common.py
+++ b/src/instructlab/eval/mt_bench_common.py
@@ -272,8 +272,7 @@ def chat_completion_openai(
             return response.choices[0].message.content
         except openai.OpenAIError as e:
             if i == API_MAX_RETRY - 1:
-                # Print error on last try
-                print(type(e), e)
+                logger.error(e)
                 break
             logger.debug(e)
             time.sleep(API_RETRY_SLEEP)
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

danmcp (approved):

danmcp (changes requested):
Until #80 is fixed, we probably need to make this be:

```
if i == max retires-1:
  logger.error
  break
logger.debug
```

Otherwise we won't give any indication of an invalid setup until the very end.

Note: This suggestion is taking the rebase into account since the break wouldn't make sense as is.

src/instructlab/eval/mt_bench_common.py
```diff
@@ -272,8 +272,7 @@ def chat_completion_openai(
             return response.choices[0].message.content
         except openai.OpenAIError as e:
             if i == API_MAX_RETRY - 1:
-                # Print error on last try
-                print(type(e), e)
+                logger.warning(e)
```
booxter, line 275: maybe even `error`? (@danmcp let me know if there's a good reason to do `info`, as in your snippet.)

booxter: I was! Will update to `error`.

nathan-weinberg (approved):
will rerun mypy check once https://github.com/instructlab/eval/pull/93 merges

danmcp:
@booxter Note I rebased your commit to pull in #93

danmcp (approved):

PR #88, opened 2024-08-01
fix: mt_bench_branch should ignore knowledge in generate

### LINKED ISSUE #87: MTBench Branch evaluator breaks when it encounters av3 taxonomy qna.yaml
It currently is trying to parse files as v2 taxonomy yaml.

bcrochet:
The mt_bench_branch currently only support pre-v3 yaml, and knowledge is currently at v3. But knowledge should not even be included in the eval run. This ignores the knowledge directory in order to skip it.

Resolves #87

### DIFF
--- a/src/instructlab/eval/mt_bench_branch_generator.py
+++ b/src/instructlab/eval/mt_bench_branch_generator.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Standard
+from pathlib import Path
 import hashlib
 import json
 import os
@@ -22,10 +23,20 @@ logger = setup_logger(__name__)
 def get_file_paths(directory):
     logger.debug(locals())
     file_paths = []
-    for root, _, files in os.walk(directory):
-        for file in files:
-            if file.split("/")[-1] == "qna.yaml":
-                file_paths.append(os.path.join(root, file))
+    root_paths = [
+        entry
+        for entry in Path(directory).iterdir()
+        if entry.is_dir()
+        if not entry.name.startswith(".")
+        if entry.name != "knowledge"
+        if entry.name != "docs"
+        if entry.name != "scripts"
+    ]
+    for basedir in root_paths:
+        for root, _, files in os.walk(basedir):
+            file_paths.extend(
+                [os.path.join(root, file) for file in files if file == "qna.yaml"]
+            )
     return file_paths
 
 


src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -23,6 +23,9 @@ def get_file_paths(directory):
     logger.debug(locals())
     file_paths = []
     for root, _, files in os.walk(directory):
+        if "knowledge" in root:
```
danmcp, line 26: This is unlikely, but this this wouldn't work if you happen to have "knowledge" in your directory structure above the taxonomy.  Or if you put "knowledge" into a skill path name.

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -23,6 +23,9 @@ def get_file_paths(directory):
...
+            logger.debug("skipping any knowledge directory")
```
danmcp, line 27: Maybe make this: Skipping knowledge directory {root}

src/instructlab/eval/mt_bench_branch_generator.py
danmcp: Something like: root.startswith(os.path.join(directory, "knowledge"))
should work

bcrochet: Good point. I'll update.

booxter: I am not sure `startswith` will work. If `directory` points to a full path, won't `root` start with `/...`?

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -23,6 +23,9 @@ def get_file_paths(directory):
     logger.debug(locals())
     file_paths = []
     for root, _, files in os.walk(directory):
+        if root.startswith(os.path.join(directory, "knowledge")):
```
booxter, line 26: I'm always confused about how `os.walk` is supposed to work cross-platform (e.g on Windows). Are we sure that `os.path.join(directory, N)` always equals to `root`? This may ring a bell: https://stackoverflow.com/questions/61252884/why-does-root-returned-from-os-walk-contain-as-directory-separator-but-os-se

I don't have access to win32 to check how this would work. I also am not sure if we care. But leaving it here just in case.

bcrochet: It looks like reworking this to use pathlib.Path might fix this? At least according to that thread...

booxter: Yes, `Path()` should probably normalize everything for you. Note that I don't think python3.11 has the native `Path.walk` so you'll have to deal with both. (first `os.walk`, then `Path(root)`...)

booxter: btw consider that the `directory`, despite some typing claims otherwise, comes from `ilab eval` where the CLI option is defined as `type=click.Path()`. I think it means that `directory` (may) arrive here as `Path`. Not sure if it affects your alternative version (if you decide to have one.)

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,7 +22,13 @@
 def get_file_paths(directory):
     logger.debug(locals())
     file_paths = []
-    for root, _, files in os.walk(directory):
+    for root, dirs, files in os.walk(directory):
+        if root == directory and "knowledge" in dirs:
+            logger.debug(
+                "skipping knowledge directory: %s", os.path.join(root, "knowledge")
+            )
+            dirs.remove("knowledge")
```
alimaredia, line 30: What's the purpose of removing "knowledge" here from `dirs`? I don't see `dirs` being used again in the function.

bcrochet: It's how os.walk is designed. As it iterates, if you remove a directory from `dirs` it does not descend into that dir. Better than going through every file in that tree and having to check its path.

alimaredia (approved):

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,7 +22,13 @@
 def get_file_paths(directory):
     logger.debug(locals())
     file_paths = []
-    for root, _, files in os.walk(directory):
+    for root, dirs, files in os.walk(directory):
```
danmcp, line 25: Recommendation: Can we switch this to an os.walk loop inside a listdir() loop for the top level?  That would avoid this root=directory check happening for the whole tree.

bcrochet: If that's the case, I feel like we should maybe move to 'just process foundational_skills and compositional_skills' and not even worry about outside of those dirs?

src/instructlab/eval/mt_bench_branch_generator.py
nathan-weinberg: Worth making this a `warning` over `debug`?

src/instructlab/eval/mt_bench_branch_generator.py
nathan-weinberg: I guess this is a question of whether or not we expect users to follow the stricter taxonomy structure we've laid out upstream or want to give them flexibility within their own taxonomy - @bjhargrave may have some helpful thoughts around this

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,10 +23,17 @@
...
+    root_path = Path(directory)
+    root_paths = [
+        entry
+        for entry in root_path.iterdir()
+        if entry.is_dir() and entry.name != "knowledge"
+    ]
+    for basedir in root_paths:
+        for root, _, files in os.walk(basedir):
+            for file in files:
+                if file.split("/")[-1] == "qna.yaml":
```
booxter, line 35: (Observation) I think `files` are already stripped of parent path, so this check should probably just directly compare `file` to `"qna.yaml"` without `.split()`.

booxter (approved):

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,10 +23,17 @@
...
     file_paths = []
-    for root, _, files in os.walk(directory):
-        for file in files:
-            if file.split("/")[-1] == "qna.yaml":
-                file_paths.append(os.path.join(root, file))
+    root_path = Path(directory)
+    root_paths = [
+        entry
+        for entry in root_path.iterdir()
+        if entry.is_dir() and entry.name != "knowledge"
```
danmcp, line 30: We still need feedback on whether the skill dirs are written in stone, but can we go ahead and filter out . directories (so we don't include .git and .github), scripts, and docs.  I don't think we will find a qna.yaml in there, but just in case

src/instructlab/eval/mt_bench_branch_generator.py
bcrochet: That's a good catch. It would just be the file name sans directory.

src/instructlab/eval/mt_bench_branch_generator.py
danmcp: ```if entry.is_dir() and not entry.name.startswith(".") and entry.name not in ["knowledge", "scripts", "docs"]```

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,10 +23,20 @@
...
     file_paths = []
-    for root, _, files in os.walk(directory):
-        for file in files:
-            if file.split("/")[-1] == "qna.yaml":
-                file_paths.append(os.path.join(root, file))
+    root_paths = [
+        entry
+        for entry in Path(directory).iterdir()
+        if entry.is_dir()
+        if not entry.name.startswith(".")
```
booxter, line 30: extracting the same attribute 4 times... CPU is cheap! :D

nathan-weinberg (approved):

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -22,10 +23,20 @@
...
+        if entry.name != "knowledge"
+        if entry.name != "docs"
+        if entry.name != "scripts"
+    ]
+    for basedir in root_paths:
+        for root, _, files in os.walk(basedir):
+            file_paths.extend(
+                [os.path.join(root, file) for file in files if file == "qna.yaml"]
```
booxter, line 38: Observation. I suspect this whole `walk`ing / `extend`ing business could be replaced with a simple `glog.glob` but I don't think we should care now, if at all.

booxter (approved):
This should merge as-is. No need to respin for my pedantry.

src/instructlab/eval/mt_bench_branch_generator.py
danmcp: Now I am curious what would be faster?  Ignoring that the compiler/interpreter will likely rewrite it anyway.  I would bet declaring a variable or an array are both slower.

booxter:
```
$ cat /tmp/test.py
from pathlib import Path
import timeit

paths = [
    Path('test')
    for i in range(10000)
]

FORBIDDEN = {"1", "2", "3"}


def test1():
    return [
        path
        for path in paths
        if not path.name.startswith('.')
        if path.name != "1"
        if path.name != "2"
        if path.name != "3"
    ]


def test2():
    return [
        path
        for path in paths
        if not path.name.startswith('.')
        if path.name not in FORBIDDEN
    ]


print(timeit.timeit(test1, number=1000))
print(timeit.timeit(test2, number=1000))
$ python /tmp/test.py
6.865049810003256
4.405094917005044
```

CPython is not very smart.

booxter: In recent pythons, we can even squash the last duplicate attribute extraction as in `test4` below:

```
$ cat /tmp/test.py
from pathlib import Path
import timeit

paths = [
    Path('test')
    for i in range(10000)
]

FORBIDDEN = {"1", "2", "3"}
FORBIDDEN_LIST = ["1", "2", "3"]


def test1():
    return [
        path
        for path in paths
        if not path.name.startswith('.')
        if path.name != "1"
        if path.name != "2"
        if path.name != "3"
    ]


def test2():
    return [
        path
        for path in paths
        if not path.name.startswith('.')
        if path.name not in FORBIDDEN
    ]


def test3():
    return [
        path
        for path in paths
        if not path.name.startswith('.')
        if path.name not in FORBIDDEN_LIST
    ]


def test4():
    return [
        path
        for path in paths
        if not (name := path.name.startswith('.'))
        if name not in FORBIDDEN
    ]


print(timeit.timeit(test1, number=1000))
print(timeit.timeit(test2, number=1000))
print(timeit.timeit(test3, number=1000))
print(timeit.timeit(test4, number=1000))
$ python /tmp/test.py
7.211299414993846
5.215823543010629
4.967952872990281
2.9443017129960936
```

-60% CPU, not too shabby. ;)

danmcp: This made my day!  Thanks for sharing.  I am surprised at the set being slower than the list (or even a statistical tie).  Normally a set has a O(1) lookup and the list would be O(n) with the set being a little more expensive to create.

Not sure what makes python attribute lookup so slow either.

Pedantic notes: I don't know if this conclusively proves which is faster still in the real life example.  The actual number of dirs is 7, only 2 of the values make it through all the ifs, and we aren't taking the set/list creation into account in this test.  Given how slow the attr lookup seems to be though, I am guessing 4) would edge out 1) with the real data/code.

Obviously the performance doesn't really matter in this case.  I would personally still pick 3 or 4 from a readability point of view.

Thanks again for putting this together!  I think it's a great practice to really understand how things are working.

booxter: @danmcp I am happy it was delightful to you too. :D If I may, I will indulge into this meaningless exercise then a bit more. :)

---

> I am surprised at the set being slower than the list (or even a statistical tie).

I think the culprit is the size of the structure. Sets are backed by hash tables, which, in a degenerate case of a short iterable, may have a similar or even higher overhead. (One has to calculate the hash function, then find the relevant bucket and perhaps even walk the bucket list.)

I [bumped](https://gist.github.com/booxter/0b16403775a131b6a88f4db1c3238df8) the size of the set of forbidden values from 3 to 100, and I get more "reasonable" numbers in terms of list vs. set(s) behavior (I also added `frozenset` to the mix - it seems to do slightly better than a regular set but not by much.)

```
if-if-if 150.47323519000201
in-set 4.468321249005385
in-list 13.755651542975102
in-frozenset 4.645498502010014
in-set-one-attr-fetch 3.846038645016961
in-list-one-attr-fetch 13.177363599010278
in-frozenset-one-attr-fetch 3.4668387069832534
```

> Not sure what makes python attribute lookup so slow either.

I think this may have something to do with the fact that `Path.name` is not a simple (or even `__slots__`) attribute but a [property](https://github.com/python/cpython/blob/fb864c76cd5e450e789a7b4095832e118cc49a39/Lib/pathlib/_local.py#L347) (meaning, a function is called on each access to the attribute). The property implementation is not huge, but it makes another `__getattr__` and a `__getitem__` calls. Perhaps a "true" attribute would fare differently. (But I have to stop falling into this rabbit hole, now!)

PR #90, opened 2024-08-01
fix: use the context correctly in mt_bench_branch

bcrochet:
The key is supposed to be context, not question.

### DIFF
--- a/src/instructlab/eval/mt_bench_branch_generator.py
+++ b/src/instructlab/eval/mt_bench_branch_generator.py
@@ -60,7 +60,7 @@ def generate(judge_model_name, branch, taxonomy_dir, output_dir):
                 if q is None or a is None:
                     continue
 
-                c = ex["question"] if "context" in ex else None
+                c = ex.get("context")
                 if c is not None:
                     t_1 = (
                         "Given the context below:\n"


alimaredia (approved):

src/instructlab/eval/mt_bench_branch_generator.py
```diff
@@ -60,7 +60,7 @@ def generate(judge_model_name, branch, taxonomy_dir, output_dir):
                 if q is None or a is None:
                     continue
 
-                c = ex["question"] if "context" in ex else None
+                c = ex["context"] if "context" in ex else None
```
booxter, line 63: nit: This would be the same as `c = ex.get("context")` I think.

danmcp: +1

danmcp: I think this is worth a quick change to avoid any future confusion as well.

bcrochet: Done.

nathan-weinberg (approved):
I will merge once the E2E run completes

PR #91, opened 2024-08-01
docs: add initial release strategy doc and CHANGELOG

nathan-weinberg:
Related to https://github.com/instructlab/dev-docs/pull/125


docs/release-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+1. Determine a commit on the main branch that will serve as the basis for the next release - most of the time this should be the latest commit.
+1. Create a new release branch in the format `release-vX.Y` off of the determined commit (will match `main` if the latest commit is chosen).
+1. Create a new release on GitHub targeting the release branch and using the latest Y-Stream tag as the previous release (e.g. `0.15.1` precedes `0.16.0`).
+1. Announce release via the following:
+    - The `#eval` channel on Slack
+    - The `dev` mailing list
+
+### Z-Stream
+
+1. Backport all relevant commits from `main` to the `release-vX.Y` branch.
```
danmcp, line 59: Noting this can also happen in the other direction.  Fix is developed for the release branch and needs to be ported to main.

Maybe state as ensure corresponding fixes are made to the release stream as well as main and any release streams between those two versions.

danmcp (changes requested):

docs/release-strategy.md
```diff
@@ -0,0 +1,83 @@
...
+
+The project currently operates on an ad-hoc release schedule based on the discretion of the maintainers team.
+
+The cadence for major releases starting from 1.0 onward will be determined as the project matures.
+
+A schedule will be updated in a markdown file on the <https://github.com/instructlab/eval> GitHub repository.
+
+## Release Tracking
+
+Currently there is no former process of release tracking. GitHub Issues are used for tracking individual work items.
```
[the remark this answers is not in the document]
nathan-weinberg: Yup, good catch

docs/release-strategy.md
nathan-weinberg: Hmmm, in CLI, all fixes go into `main` first and then are backported. When would the reverse occur?

danmcp: It depends on the driver, if we find the issue on main, we would typically backport as needed.  If the issue is reported by a user on a particular version, you might recreate and fix it there first to get that user enabled.  Especially if the work was very different between the two versions.

nathan-weinberg: I feel like if update that here, we should do the same in the CLI. AFAIK we are only supporting the most recent Y-Stream at this juncture. So we can do that update in both, but I'm not sure if it'll come into play until we start supporting multiple versions.

danmcp: For our v0.1.2 release, I ended up merging to the release branch first:

https://github.com/instructlab/eval/pull/110

and then backporting to main:

https://github.com/instructlab/eval/pull/112

In this case I did it because we wanted to get the fix out asap and didn't want to wait for tests on main, then backport and wait for test on the release branch.

nathan-weinberg: Very good point - I'm going to update the doc with some wording for that - once that's in I think we could go for one last round of reviews and merge this

docs/release-strategy.md
```diff
@@ -0,0 +1,84 @@
...
+1. Create a new release branch in the format `release-vX.Y` off of the determined commit (will match `main` if the latest commit is chosen).
+1. Create a new release on GitHub targeting the release branch and using the latest Y-Stream tag as the previous release (e.g. `0.15.1` precedes `0.16.0`).
+1. Announce release via the following:
+    - The `#eval` channel on Slack
+    - The `dev` mailing list
+
+### Z-Stream
+
+1. Backport all relevant commits from `main` to the `release-vX.Y` branch.
+    - It may also be the case you wish to push commits directly to the release branch - if this approach is taken, ensure any relevant commits are subseqently backported to `main`
```
danmcp, line 60: Maybe state this as:

It may also be the case you wish to update the release branch first - ...

Or similar.  Probably don't want to encourage direct pushing to branches.

danmcp (commented):
Note the spellcheck linter as well

docs/release-strategy.md
nathan-weinberg: Updated the wording and fixed the spellchecker

PR #93, opened 2024-08-05
github: add action to free runner disk space for tox installs

nathan-weinberg:
Same as what we are using in https://github.com/instructlab/instructlab


booxter (changes requested):
You will have to `uses: ./.github/actions/free-disk-space` it in the relevant workflows.

booxter (approved):
It looks correct. Of course, the final say is for CI job to succeed...

danmcp (approved):

alimaredia (approved):

PR #98, opened 2024-08-13
Remove fastchat dependency

danmcp:
There were only a couple of remaining calls to fastchat.  Models we don't interact with have been removed from support.

pylint version has been updated to correspond to cli and to allow the type | None syntax.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
-FastChat
 GitPython>=3.1.42,<4.0.0
 shortuuid
 openai>=1.13.3,<2.0.0
--- a/src/instructlab/eval/mt_bench_answers.py
+++ b/src/instructlab/eval/mt_bench_answers.py
@@ -6,8 +6,6 @@ import os
 import time
 
 # Third Party
-# TODO need to look into this dependency
-from fastchat.model.model_adapter import get_conversation_template  # type: ignore
 import shortuuid
 import tqdm
 
@@ -20,6 +18,7 @@ from .mt_bench_common import (
     load_questions,
     temperature_config,
 )
+from .mt_bench_model_adapter import get_conversation_template  # type: ignore
 
 logger = setup_logger(__name__)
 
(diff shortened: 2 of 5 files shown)


danmcp:
@alimaredia FYI

danmcp:
@dhellmann Note this will remove the fastchat dependency

dhellmann:
> @dhellmann Note this will remove the fastchat dependency

Ack, thanks for the heads-up.

jaideepr97:
@danmcp are we confident we don't need to rely on fastchat for any future developments to the code that we are importing from them? 
This seems great from a dependency reduction point of view, but we would essentially be "freezing" the borrowed code (unless we plan to develop it in a different direction)

nathan-weinberg (approved):

danmcp:
> @danmcp are we confident we don't need to rely on fastchat for any future developments to the code that we are importing from them? 

I wish I could predict the future and answer this question :). We had already included the bulk of the fastchat code we needed within eval.  This change largely includes the remainder of customizations from xukai92 and leaves our support for most model type conversations.

https://github.com/lm-sys/FastChat/compare/main...xukai92:FastChat:ilab

that aren't included in FastChat anyway.  Given the amount of customization we are doing already, and the amount the code is tied to OpenAI vs FastChat, I think we are stuck maintaining our own version that works with the versions of OpenAI and vllm and models we need to support.

I think the biggest risk comes to changes we need to make to the evaluation conversation with the model and judge and less to the core code.


> This seems great from a dependency reduction point of view, but I'm guessing we would essentially be "freezing" the borrowed code (unless we plan to develop it in a different direction)

There is already a decent bit of divergence with Kai's changes, adding support for a new OpenAI version, support for custom judges, and support for merge-system-user-message.

This is definitely a good question to ask.  This particular PR doesn't increase the risk very much vs what we had already done, but the overall problem of how to maintain everything long term still exists.

jaideepr97 (approved):

alinaryan (approved):

khaledsulayman (approved):

alimaredia (approved):

PR #99, opened 2024-08-14
Allow for external serving to be used with mmlu

### LINKED ISSUE #50: Enable MMLU tests to be run on an already served model
Right now the inferencing done during MMLU is done within the lm-eval-harness library. Lm-eval-harness has the ability to inference with openai-api compatible servers similar to how models are being served via `ilab serve`:

https://github.com/EleutherAI/lm-evaluation-harness?tab=readme-ov-file#model-apis-and-inference-servers

A fix for this issue would involve serving a model on an endpoint with either vLLM or llama-cpp and being able to pass that endpoint into `MMLUEvaluator.run()` or the `MMLUEvaluator` class to run MMLU tests.

### LINKED ISSUE #68: mmlu isn't consuming multiple gpus
Steps to recreate:

Launch mmlu on an instance with multiple gpus.  Run:

ilab model evaluate --model models/instructlab/granite-7b-lab --benchmark mmlu

Only 1 gpu is consumed.  Adjusting batch-size doesn't seem to have any effect.

danmcp:
Requires lm_eval >= 0.4.4

As implemented, if server_url is passed, it's used for external serving.  Otherwise the old logic is still applied.  Eventually we may want to remove the non external serving option.

Resolves: #50
Resolves: #68

Corresponding cli change: https://github.com/instructlab/instructlab/pull/2065

### DIFF
--- a/src/instructlab/eval/mmlu.py
+++ b/src/instructlab/eval/mmlu.py
@@ -122,8 +122,14 @@ class AbstractMMLUEvaluator(Evaluator):
         self.batch_size = batch_size
         self.device = device
 
-    def _run_mmlu(self) -> dict:
-        model_args = f"pretrained={self.model_path},dtype={self.model_dtype}"
+    def _run_mmlu(self, server_url: str | None = None) -> dict:
+        if server_url is not None:
+            # Requires lm_eval >= 0.4.4
+            model_args = f"base_url={server_url}/completions,model={self.model_path},tokenizer_backend=huggingface"
+            model = "local-completions"
+        else:
+            model_args = f"pretrained={self.model_path},dtype={self.model_dtype}"
+            model = "hf"
         tm = None
         if self.tasks_dir is not None:
             if not os.path.exists(self.tasks_dir):
@@ -132,7 +138,7 @@ class AbstractMMLUEvaluator(Evaluator):
                 raise InvalidTasksDirError(self.tasks_dir)
             tm = TaskManager(verbosity="DEBUG", include_path=self.tasks_dir)
         mmlu_output = self._simple_evaluate_with_error_handling(
-            model="hf",
+            model=model,
             model_args=model_args,
             tasks=self.tasks,
             num_fewshot=self.few_shots,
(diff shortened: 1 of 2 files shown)


danmcp:
Note: Tests will fail until lm_eval 0.4.4 is available

danmcp:
@JamesKunstle FYI

src/instructlab/eval/mmlu.py
```diff
@@ -122,8 +122,14 @@ def __init__(
         self.batch_size = batch_size
         self.device = device
 
-    def _run_mmlu(self) -> dict:
-        model_args = f"pretrained={self.model_path},dtype={self.model_dtype}"
+    def _run_mmlu(self, server_url: str | None = None) -> dict:
+        if server_url is not None:
```
jaideepr97, line 126: should we do some kind of basic validation on the provided `server_url` before using it?

jaideepr97:
cool patch btw! nice way to get around the current limitations 
could you take a look at the CI failures?

src/instructlab/eval/mmlu.py
danmcp: Four answers:
- No because it's just a passthrough.  Can let the actual consumer (lm_eval) do the validation.
- Yes because we are adding /completions to it. This is required as of lm_eval 0.4.4.
- No because we aren't validating server_url anywhere else.
- Yes, but maybe we should do it in a follow-up and do it consistently across all of eval and maybe instructlab

danmcp:
> cool patch btw! nice way to get around the current limitations could you take a look at the CI failures?

The CI failures are because of requiring lm_eval 0.4.4.  They are promising a new release should come out soon.  We're in a holding pattern until then.

src/instructlab/eval/mmlu.py
jaideepr97: I think I agree with point 4, that kind of automatically addresses all other points

danmcp:
@mergifyio rebase

alimaredia (approved):

nathan-weinberg (approved):

nathan-weinberg:
https://github.com/EleutherAI/lm-evaluation-harness/releases/tag/v0.4.4 has been released

jaideepr97 (approved):

Issue #97, opened 2024-08-13
Auto detect mixtral type judge models and auto set merge_system_user_message


danmcp:
mt_bench currently take an option for merge_system_user_message which is needed for mixtral type judge models (Ex: prometheus).  If mixtral can be auto detected, the need for this setting can be removed.

An alternate methodology might be to determine if the model is mixtral based on the format for the results. And if the format of the result looks like it was from mixtral, assume it was a mixtral model and parse it accordingly.

PR #100, opened 2024-08-15
Default to merge_system_user_message if mistral model detected

danmcp:
Resolves: https://github.com/instructlab/eval/issues/97

Requires https://github.com/instructlab/eval/pull/98 to be able to function but won't cause any harm to merge before.

### DIFF
--- a/src/instructlab/eval/mt_bench_common.py
+++ b/src/instructlab/eval/mt_bench_common.py
@@ -254,7 +254,7 @@ def chat_completion_openai(
         try:
             messages = conv.to_openai_api_messages()
             if (
-                merge_system_user_message
+                (merge_system_user_message or conv.name == "mistral")
                 and messages[0]["role"] == "system"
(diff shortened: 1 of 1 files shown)


danmcp:
@JamesKunstle FYI

src/instructlab/eval/mt_bench_common.py
```diff
@@ -254,7 +254,7 @@ def chat_completion_openai(
         try:
             messages = conv.to_openai_api_messages()
             if (
-                merge_system_user_message
+                (merge_system_user_message or conv.name == "mistral")
```
JamesKunstle, line 257: definitely solved for mistral (and mixtral). I wonder if there are other models that disallow system messages. If so, we could enum the names and extend that list. This patch is perfect as is though- just food for thought.

JamesKunstle (approved):

nathan-weinberg (approved):

Issue #95, opened 2024-08-12
Request: max_workers auto tuning


danmcp:
Currently, max-workers defaults to 16 and gives a recommendation when run on how to tune based on hardware from the cli.  This isn't ideal since the recommendation is given after the run is started.  Also, training calls evaluate and needs and needs a more programatic way to tune.  This request is to give an 'auto' option for max workers that will give it a an appropriate value based on the available hardware configuration.

Related: https://github.com/instructlab/instructlab/issues/2050

PR #101, opened 2024-08-15
Add optional auto tuning for max_workers

danmcp:
To accomplish this, serving_gpus will also need to be passed by the caller.  This logic also removes the magic number if max_workers is not passed and relies on the cpu count instead.

Resolves: https://github.com/instructlab/eval/issues/95

More info on timings that influenced this algorithm: https://github.com/instructlab/instructlab/issues/2050

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -39,6 +39,21 @@ class InvalidModelError(EvalError):
         self.message = f"Model found at {path} but was invalid due to: {reason}"
 
 
+class InvalidMaxWorkersError(EvalError):
+    """
+    Error raised when max_workers isn't an int or "auto"
+
+    Attributes
+        message         error message to be printed on raise
+        max_workers     max_workers specified
+    """
+
+    def __init__(self, max_workers) -> None:
+        super().__init__()
+        self.max_workers = max_workers
+        self.message = f"Invalid max_workers '{max_workers}' specified. Valid values are positive integers or 'auto'."
+
+
 class InvalidGitRepoError(EvalError):
     """
     Error raised when taxonomy dir provided isn't a valid git repo
--- a/src/instructlab/eval/mt_bench.py
+++ b/src/instructlab/eval/mt_bench.py
@@ -6,12 +6,17 @@ https://klu.ai/glossary/mt-bench-eval
 https://arxiv.org/html/2306.05685
 """
 
+# Standard
+import multiprocessing
+import os
+
 # First Party
 from instructlab.eval import (
     mt_bench_answers,
     mt_bench_branch_generator,
     mt_bench_judgment,
 )
+from instructlab.eval.exceptions import InvalidMaxWorkersError
 
 # Local
 from .evaluator import Evaluator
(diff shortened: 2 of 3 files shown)


danmcp:
@JamesKunstle FYI

src/instructlab/eval/mt_bench.py
```diff
@@ -20,34 +24,71 @@
...
         self.output_dir = output_dir
-        self.max_workers = max_workers
+        self.serving_gpus = serving_gpus
         self.merge_system_user_message = merge_system_user_message
 
+        if max_workers == "auto":
+            if serving_gpus is not None:
+                # Tune max_workers based on hardware configuration: min(#GPUs being used * 10, #CPU cores)
+                self.max_workers = min(
+                    max(serving_gpus, 1) * 10, multiprocessing.cpu_count()
```
jaideepr97, line 59: looking at https://docs.python.org/3/library/multiprocessing.html#multiprocessing.cpu_count should we be using `len(os.sched_getaffinity(0))` instead? 

also if you don't mind elaborating on how you arrived at this expression to determine the # of max workers that would be interesting to me

src/instructlab/eval/mt_bench.py
```diff
@@ -20,34 +24,71 @@
...
+                # Don't be too aggressive when serving_gpus isn't specified. Use half the cpu count.
+                self.max_workers = multiprocessing.cpu_count() // 2
+                logger.debug(
+                    "max_workers set to auto but serving_gpus is not specified. Defaulting to (cpu count / 2): %s",
+                    self.max_workers,
+                )
+        else:
+            if isinstance(max_workers, int):
+                logger.debug("max_workers specified as: %s", max_workers)
+                self.max_workers = max_workers
```
jaideepr97, line 72: should we check to make sure `max_workers` > 0 here?

src/instructlab/eval/exceptions.py
```diff
@@ -39,6 +39,21 @@ def __init__(self, path, reason) -> None:
...
+    Error raised when max_workers isn't an int or "auto"
+
+    Attributes
+        message         error message to be printed on raise
+        max_workers     max_workers specified
+    """
+
+    def __init__(self, max_workers) -> None:
+        super().__init__()
+        self.max_workers = max_workers
```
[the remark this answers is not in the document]
danmcp: It is purely future facing.  And consistent with the other exception impls.

src/instructlab/eval/mt_bench.py
danmcp: > also if you don't mind elaborating on how you arrived at this expression to determine the # of max workers that would be interesting to me

https://github.com/instructlab/instructlab/issues/2050

danmcp: > should we be using `len(os.sched_getaffinity(0))` instead?

Yes! Will address

src/instructlab/eval/mt_bench.py
danmcp: The check wasn't there before, but technically yes.  Will address.

src/instructlab/eval/mt_bench.py
danmcp: Fixed to use len(os.sched_getaffinity(0))

nathan-weinberg (approved):

jaideepr97 (approved):

jaideepr97:
added hold because i think we might want to merge https://github.com/instructlab/instructlab/pull/2072 first? 
and possibly give @danmcp a chance to address the one minor comment I have open

danmcp:
> added hold because i think we might want to merge [instructlab/instructlab#2072](https://github.com/instructlab/instructlab/pull/2072) first?

They can be merged in either order.  I will need to do followups to use the "auto" feature from the cli for eval and training once eval is released. 


> and possibly give @danmcp a chance to address the one minor comment I have open if he chooses to

Added the comment.

PR #103, opened 2024-08-16
Dont retry on connection failure

danmcp:
This is a continuation/replacement of https://github.com/instructlab/eval/pull/80

Resolves: #77

The approach of this separates errors into 3 buckets:

Fatal errors we retry and fail after max attempts (Ex: APIConnectionError, 404).  These cases don't have much hope of fixing themselves.
Errors which we record after max attempts (Ex: RateLimitError).  Failure on one try might not mean they are all going to fail.
Errors which we record after 1 occurrence (Ex: 400). Failure is request specific and retries aren't going to help.

Which allows the logic to fail fast when possible, not retry when it wouldn't help, and not catastrophically fail on isolated failures.

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -107,3 +107,15 @@ class InvalidTasksDirError(EvalError):
         super().__init__()
         self.tasks_dir = tasks_dir
         self.message = f"Invalid Tasks Dir: {tasks_dir}"
+
+
+class OpenAIError(EvalError):
+    """
+    Error raised when reply retrieval from OpenAI API fails.
+    Attributes
+        message              error message to be printed on raise
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.message = "Failed to receive a reply from API."
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -109,13 +109,13 @@ class InvalidTasksDirError(EvalError):
         self.message = f"Invalid Tasks Dir: {tasks_dir}"
 
 
-class OpenAIError(EvalError):
+class ModelServingAPIError(EvalError):
     """
-    Error raised when reply retrieval from OpenAI API fails.
+    Error raised when reply retrieval from model serving fails.
     Attributes
         message              error message to be printed on raise
     """
 
     def __init__(self) -> None:
         super().__init__()
-        self.message = "Failed to receive a reply from API."
+        self.message = "Failed to receive a reply from model serving API."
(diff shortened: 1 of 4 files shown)


src/instructlab/eval/mt_bench_common.py
```diff
@@ -247,37 +251,99 @@ def play_a_match_single(
...
+        (
+            openai.APIConnectionError,
+            openai.AuthenticationError,
+            openai.PermissionDeniedError,
+            openai.NotFoundError,
+        ),
+    )
+
+
+# TODO: copied from instructlab (cli) utils module; consolidate somewhere?
```
jaideepr97, line 266: can we just import this from the cli? if we've made sure to keep our imports tight everywhere we may not have to worry about import cycles in that case

src/instructlab/eval/exceptions.py
```diff
@@ -107,3 +107,15 @@ def __init__(self, tasks_dir) -> None:
...
+class ModelServingAPIError(EvalError):
+    """
+    Error raised when reply retrieval from model serving fails.
+    Attributes
+        message              error message to be printed on raise
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.message = "Failed to receive a reply from API."
```
jaideepr97, line 121:
```suggestion
        self.message = "Failed to receive a reply from model serving API."
```

jaideepr97 (approved):
Thanks @danmcp 
I really like how you've broken down the errors by error code - it helps to further my understanding of good error handling in general 

we could probably start focusing on expanding the unit testing infrastructure on the project but that should probably be in a separate PR

src/instructlab/eval/mt_bench_common.py
jaideepr97: oh yeah, you're right

nathan-weinberg (approved):

src/instructlab/eval/mt_bench_common.py
```diff
@@ -247,37 +251,99 @@ def play_a_match_single(
...
+            openai.APIConnectionError,
+            openai.RateLimitError,  # 429
+            openai.InternalServerError,  # >=500
+            # NOTE: Errors listed below may need a revisit: we are not sure if
+            # it's ever helpful to retry them. Leaving them intact for now.
+            openai.AuthenticationError,  # 401
+            openai.PermissionDeniedError,  # 403
+            openai.NotFoundError,  # 404
+            # General catch-all
+            openai.OpenAIError,
```
booxter, line 331: So AFAIU all the explicit types listed above are for documentation purposes only? Because all of them are OpenAIErrors already.

booxter (commented):
This is fine now. Not sure I can approve it since most of the PR is (co)authored by myself.

src/instructlab/eval/mt_bench_common.py
danmcp: Yes.  I thought they were really helpful to see elaborated.  Although it's a great point that we should add a comment stating that logic explicitly.

PR #105, opened 2024-08-21
Add comment to make it clear how the code is working

danmcp:
The logic isn't technically necessary but it seems helpful to list the exceptions we expect to fit into this bucket.


nathan-weinberg (approved):

nathan-weinberg:
@mergifyio refresh

alinaryan (approved):

nathan-weinberg:

PR #106, opened 2024-08-22
Better path and string handling

danmcp:
Removing several leftover cases of paths formed without os.path.join
Replacing string concats with f""

### DIFF
--- a/src/instructlab/eval/mt_bench_judgment.py
+++ b/src/instructlab/eval/mt_bench_judgment.py
@@ -169,12 +169,12 @@ def judge_model(
     data_base_dir = bench_dir(data_dir, bench_name, branch)
     output_base_dir = bench_dir(output_dir, bench_name, branch)
 
-    judge_file = f"{package_data_dir}/{bench_name}/judge_prompts.jsonl"
+    judge_file = os.path.join(package_data_dir, bench_name, "judge_prompts.jsonl")
 
-    question_file = f"{data_base_dir}/question.jsonl"
-    answer_file = f"{output_base_dir}/model_answer/{model_name}.jsonl"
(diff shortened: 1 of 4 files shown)


nathan-weinberg (approved):

src/instructlab/eval/mt_bench_judgment.py
```diff
@@ -195,7 +195,9 @@ def judge_model(
         models = model_list
 
     judges = make_judge_single(judge_model_name, judge_prompts)
-    output_file = f"{output_base_dir}/model_judgment/{judge_model_name}_single.jsonl"
+    output_file = os.path.join(
+        output_base_dir, "model_judgment", judge_model_name, "_single.jsonl"
```
JamesKunstle, line 199: I think this should be:
```suggestion
        output_base_dir, "model_judgment", f"{judge_model_name}_single.jsonl"
```
Otherwise `/_single.jsonl` is a dile in the `judge_model_name` dir. Maybe this is your intention though

JamesKunstle:
pathlib.Path is the string-handling mechanism in the CLI, would there be any interest in switching over for this library?

src/instructlab/eval/mt_bench_judgment.py
danmcp: Fixed, thanks for catching!

danmcp:
> pathlib.Path is the string-handling mechanism in the CLI, would there be any interest in switching over for this library?

I would generally prefer pathlib.Path but I was trying to make the simplest change.  We also aren't really doing any operations here where the niceties of pathlib.Path would come in handy.

nathan-weinberg:
@mergifyio refresh

jaideepr97 (approved):

nathan-weinberg:

PR #107, opened 2024-08-23
Allow max_workers to be passed in after evaluator is created

danmcp:
The reason this is necessary is serving_gpus might not be available when the evaluator is constructed.  A typical flow might be:

- Create evaluator
- Launch server # This is when serving_gpus is probably calculated since it will depend on the model being served
- generate
- judge

The new logic allows for max_workers and serving_gpus to be passed in as the generate and judge steps occur.

Once all the known callers have been updated (eval and training) we can remove the two attrs from the constructors and make the logic a little simpler.

Corresponding cli change: https://github.com/instructlab/instructlab/pull/2144
This change will need to be released first.

### DIFF
--- a/src/instructlab/eval/mt_bench.py
+++ b/src/instructlab/eval/mt_bench.py
@@ -52,31 +52,49 @@ class AbstractMTBenchEvaluator(Evaluator):
         self.output_dir = output_dir
         self.serving_gpus = serving_gpus
         self.merge_system_user_message = merge_system_user_message
+        self.max_workers = self._calc_max_workers(max_workers, serving_gpus)
 
-        if max_workers == "auto":
-            try:
-                # Not available on all platforms
-                usable_cpu_count = len(os.sched_getaffinity(0))  # type: ignore[attr-defined]
-            except AttributeError:
-                usable_cpu_count = multiprocessing.cpu_count()
-            if serving_gpus is not None:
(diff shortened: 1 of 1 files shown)


src/instructlab/eval/mt_bench.py
```diff
@@ -52,31 +52,49 @@ def __init__(
...
+        self, max_workers: int | str | None, serving_gpus: int | None
+    ):
+        calculated_max_workers = None
+        if max_workers is not None:
+            if max_workers == "auto":
+                try:
+                    # Not available on all platforms
+                    usable_cpu_count = len(os.sched_getaffinity(0))  # type: ignore[attr-defined]
+                except AttributeError:
+                    usable_cpu_count = multiprocessing.cpu_count()
```
nathan-weinberg, line 67: What is the advantage of `os.sched_getaffinity(0)` over `multiprocessing.cpu_count()`?

nathan-weinberg (approved):
One question but otherwise LGTM - thanks @danmcp!

src/instructlab/eval/mt_bench.py
nathan-weinberg: Makes sense, is it worth mentioning that in a code comment? Either way won't block on this, appreciate the explanation!

alinaryan (approved):

PR #110, opened 2024-08-27
Use single answer file and model list

danmcp:
Since phased training is using individual model names for judgment, it's hitting the case where multiple models are being loaded for evaluation.  We don't currently need that code path as it was inherited from FastChat so this change removes it and relies on the single answerfile and its model type.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Run e2e test
         run: |
           . venv/bin/activate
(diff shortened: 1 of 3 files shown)


jaideepr97 (approved):
LGTM thanks @danmcp

nathan-weinberg (approved):

JamesKunstle:
Fixes https://github.com/instructlab/instructlab/issues/2148

danmcp:
@mergifyio backport main

PR #111, opened 2024-08-27
Improve logging

danmcp:
This removes an unused dict (which was previously used for logging) and replaces it with logger.debug statements to avoid the extra object creation for logging.


nathan-weinberg (approved):

nathan-weinberg:
@mergifyio refresh

alinaryan (approved):

nathan-weinberg:

PR #115, opened 2024-08-28
Cleanup usage of load model answers

danmcp:
This logic removes the directory walking logic of load_model_answers in favor of finding the exact file needed.  Since we are no longer attempting to support multiple models within a single eval, this simplification is possible and avoids some confusing logic that was in place to allow any judge model to be passed but still use the builtin reference answers.

### DIFF
--- a/src/instructlab/eval/mt_bench_common.py
+++ b/src/instructlab/eval/mt_bench_common.py
@@ -87,32 +87,15 @@ def load_questions(question_file: str, begin: Optional[int], end: Optional[int])
     return questions
 
 
-def load_model_answers(answer_dir: str, model_name=None, answer_file=None) -> dict:
-    """Load model answers.
+def load_model_answers(answer_file: str, model_name: str | None = None) -> dict:
+    """Load model answers from a single answer file
 
     The return value is a python dict of type:
     Dict[model_name: str -> Dict[question_id: int -> answer: dict]]
     """
     logger.debug(locals())
-    model_answers = {}
-    if answer_file is not None:
-        filename = os.path.basename(answer_file)
-        # Removing ".jsonl"
-        file_model_name = filename[:-6]
-        model_answers[file_model_name] = _load_answers(answer_file)
-    else:
-        for root, _, files in os.walk(answer_dir):
-            for filename in files:
-                if filename.endswith(".jsonl"):
-                    # Removing ".jsonl"
-                    file_model_name = filename[:-6]
-                    file_path = os.path.join(root, filename)
-                    model_answers[model_name or file_model_name] = _load_answers(
-                        file_path
-                    )
-                    if model_name == file_model_name:
-                        logger.debug("Found answer file matching: %s", model_name)
-                        break
+    file_model_name = os.path.splitext(os.path.basename(answer_file))[0]
+    model_answers = {model_name or file_model_name: _load_answers(answer_file)}
     return model_answers
 
 
--- a/src/instructlab/eval/mt_bench_judgment.py
+++ b/src/instructlab/eval/mt_bench_judgment.py
@@ -162,7 +162,9 @@ def judge_model(
     """Judge the model based on questions and reference answers"""
     logger.debug(locals())
     package_data_dir = os.path.join(os.path.dirname(__file__), "data")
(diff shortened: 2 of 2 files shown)


src/instructlab/eval/mt_bench_judgment.py
```diff
@@ -172,15 +174,19 @@ def judge_model(
 
     question_file = f"{data_base_dir}/question.jsonl"
     answer_file = f"{output_base_dir}/model_answer/{model_name}.jsonl"
-    answer_dir = os.path.dirname(answer_file)
-    ref_answer_dir = f"{data_base_dir}/reference_answer"
+    if use_builtin_ref_answers:
+        ref_answer_file = os.path.join(data_base_dir, "reference_answer", "gpt-4.jsonl")
```
nathan-weinberg, line 178: I assume GPT-4 is the default format here due to lm_eval?

src/instructlab/eval/mt_bench_common.py
```diff
@@ -87,32 +87,15 @@ def load_questions(question_file: str, begin: Optional[int], end: Optional[int])
...
-                    file_model_name = filename[:-6]
-                    file_path = os.path.join(root, filename)
-                    model_answers[model_name or file_model_name] = _load_answers(
-                        file_path
-                    )
-                    if model_name == file_model_name:
-                        logger.debug("Found answer file matching: %s", model_name)
-                        break
+    file_model_name = os.path.splitext(os.path.basename(answer_file))[0]
+    model_answers = {model_name or file_model_name: _load_answers(answer_file)}
```
nathan-weinberg, line 98: I'm not familiar with this Python syntax - is this like a lambda? i.e. if either `model_name` or `file_model_name` evaluates to `true` then `_load_answers` is called?

src/instructlab/eval/mt_bench_common.py
```diff
@@ -87,32 +87,15 @@ def load_questions(question_file: str, begin: Optional[int], end: Optional[int])
     return questions
 
 
-def load_model_answers(answer_dir: str, model_name=None, answer_file=None) -> dict:
-    """Load model answers.
+def load_model_answers(answer_file, model_name=None) -> dict:
```
nathan-weinberg, line 90: We've lost the type hint here - could this be multiple different types? It seems from the `_load_answers` func definition this always needs to be a JSON file: https://github.com/danmcp/eval/blob/62e4bc3e5dee05958cab2182f40cd8b6df15cbec/src/instructlab/eval/mt_bench_common.py#L102

src/instructlab/eval/mt_bench_judgment.py
danmcp: gpt4.jsonl are the current committed reference answers for mt_bench: https://github.com/instructlab/eval/blob/main/src/instructlab/eval/data/mt_bench/reference_answer/gpt-4.jsonl

nathan-weinberg: Ack, thanks for clarifying that for me!

src/instructlab/eval/mt_bench_common.py
danmcp: It's the equivalent of writing:

```
if model_name:
  keyname = model_name
else
  keyname = file_model_name
model_answers = {keyname: _load_answers(answer_file)}
```

So if model_name isn't None, it will satisfy the or and resolve to its value.

src/instructlab/eval/mt_bench_common.py
danmcp: answer_file didn't have a type hint before (answer_dir did), but happy to add one.

src/instructlab/eval/mt_bench_common.py
nathan-weinberg: Very cool, thanks for the explanation!

src/instructlab/eval/mt_bench_common.py
nathan-weinberg: Ah you're right, thanks for the addition!

nathan-weinberg (approved):

jaideepr97 (approved):

PR #135, opened 2024-09-24
Allow passing a device to the eval benchmarks

alinaryan:


alinaryan:
Closing since https://github.com/instructlab/instructlab/pull/2065 will be extracting serve from mmlu, meaning the device handling is managed by the external server.

PR #138, opened 2024-09-26
return overall_score from MTBenchBranch.generate_judgement()

alimaredia:
This allows the overall_score to be shown by
callers of the library along with qa pairs
and the error rate.

### DIFF
--- a/src/instructlab/eval/mt_bench.py
+++ b/src/instructlab/eval/mt_bench.py
@@ -260,10 +260,12 @@ class MTBenchBranchEvaluator(AbstractMTBenchEvaluator):
             serving_gpus    Number of gpus allocated for serving.  Used to tune with max_workers=auto.  None indicates to use value specified in constructor.
 
         Returns:
+            overall_score   overall score from the evaluation
             qa_pairs        Question and answer pairs (with scores) from the evaluation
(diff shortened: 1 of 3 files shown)


danmcp:
@sallyom FYI on the api change

danmcp:
Note: This can't merge until a change is made to the cli to allow for the 2 and 3 param tuple return at the same time.

danmcp (approved):
Looks good, just need to fix the formatting error

nathan-weinberg (approved):

alinaryan (approved):

src/instructlab/eval/mt_bench.py
```diff
@@ -260,10 +260,12 @@ def judge_answers(
             serving_gpus    Number of gpus allocated for serving.  Used to tune with max_workers=auto.  None indicates to use value specified in constructor.
 
         Returns:
+            overall_score   overall score from the evaluation
             qa_pairs        Question and answer pairs (with scores) from the evaluation
+            error_rate      percentage of questions dropped due to errors during evaluation
```
danmcp, line 265: If you do make another change to this commit, this is also missing for the mt_bench case above

PR #140, opened 2024-09-26
Remove max_workers and serving_gpus from constructor

danmcp:
Removes max_workers from the constructor in favor of passing to judge_answers and gen_answers.  The original change was made because workers should be dependent on the particular settings of the model being served which can vary between the model being judged and the judge itself.

Any callers to max_workers= or serving_gpus= in the constructor should switch to use gen_answers and judge_answers for MTBenchEvaluator and MTBenchBranchEvaluator.

Original change made here: https://github.com/instructlab/eval/pull/107
CLI updated here: https://github.com/instructlab/instructlab/pull/2144

This change is not backward compatible with callers using max_workers or serving_gpus in the constructor.  A minor version bump should be made for the next release.

### DIFF
--- a/src/instructlab/eval/mt_bench.py
+++ b/src/instructlab/eval/mt_bench.py
@@ -33,8 +33,6 @@ class AbstractMTBenchEvaluator(Evaluator):
         model_name                  Name of the model to evaluate
         judge_model_name            Name of the judge model
         output_dir                  The directory to use for evaluation output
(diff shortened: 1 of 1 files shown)


danmcp:
@sallyom FYI

alinaryan (approved):

nathan-weinberg (approved):

PR #149, opened 2024-10-11
Add traceback for InvalidTasksDirError

alimaredia:
There are two instance of `InvalidTasksDirError` being thrown. Without a stack trace it can be difficult to determine which instance was thrown.

### DIFF
--- a/src/instructlab/eval/exceptions.py
+++ b/src/instructlab/eval/exceptions.py
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
+# Standard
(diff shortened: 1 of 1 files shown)


nathan-weinberg (approved):

danmcp (changes requested):
To not put a stack trace in an exception message, favoring the change being worked on here:
https://github.com/instructlab/instructlab/pull/2432/files

alimaredia:
Closed in favor of: https://github.com/instructlab/instructlab/pull/2432

PR #155, opened 2024-10-21
ci: change small E2E CI job to medium

nathan-weinberg:
the small E2E CI job doesn't actually test evaluation
change to a job we will get some use out of

resolves https://github.com/instructlab/eval/issues/12


nathan-weinberg:
@danmcp we'll need the `HF_TOKEN` secret here as well - I can add a temp token at the repo-level if needed

danmcp (approved):

PR #156, opened 2024-10-22
Unitxt evaluator

### LINKED ISSUE #176: Evaluation of user data using Unitxt
Here is the suggested flow. Let's discuss in a meeting to see it makes sense and modify as needed: 

Evaluation command `[ilab model evaluate new_data`] will have the following parameters:

1 - `csv_path` for user data

- 	A csv file with two required column ('instruction','input') and two optional columns ('answer','context')
- 	'context' column is for the RAG task only
- 	'answer' is the golden truth, if available
- 	'instruction' explain the task ("Summarize this text", "Complete this sentence", "Classify this input to one of the following: ... ")

	
2 - `task_type` out of the following options:

- 	Classification
- 	Question Answering [Let's discuss if we want to explicitly offer QA multichoice and simple QA as two separate options]
- 	Summarization
- 	Generation
- 	RAG
- 	Other [Let's discuss if this can actually be removed as it will get the same treatment as QA behind the scenes]

3 - `use_llmaaj` (False by default)

- 	False - default standard metric for the task 
    - for some tasks the default is llmaaj to begin with: QA, Generation, Other
    - llmaaj will be used if golden answers are not available
- 	True - Uses judge (with templates pre defined for the task type)

4 - `num_shots` (0 by default)
-   Do we want to allow the user select num shots?
-   Do we want to drop this option, run a few configurations (0, 2, 5 shots) and inform the user which is the best setting?
 

Following the command, unitxt will  run the provided data with the task of choice, and replace the metric if llmaaj is selected.
The data will be run in multiple configurations (fitted into different templates that match the task).
Results will include a recommendation for the best template of those used.

Roni-Friedman:
Adding [unitxt](https://github.com/IBM/unitxt) evaluator.

To be complemented by [adding unitxt as a benchmark in instructlab repo](https://github.com/instructlab/instructlab/pull/2517)


nathan-weinberg:
@Mergifyio refresh

tests/test_unitxt.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_unitxt():
+    print("===> Executing 'test_unitxt'...")
+    try:
+        model_path = "instructlab/granite-7b-lab"
+        unitxt_recipe = "card=cards.wnli,template=templates.classification.multi_class.relation.default,max_train_instances=5,loader_limit=20,num_demos=3,demos_pool_size=10"
+        unitxt = UnitxtEvaluator(model_path=model_path, unitxt_recipe=unitxt_recipe)
+        overall_score, single_scores = unitxt.run()
+        print(overall_score)
```
yoavkatz, line 12: I think the unitest should check the overall_score and see that they make sense.  This will ensure the api does not break at some point (E.g .returns 'none').

src/instructlab/eval/unitxt.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+
+    def remove_unitxt_files(self):
+        if self.tasks_dir.startswith(
+            TEMP_DIR_PREFIX
+        ):  # to avoid unintended deletion if this class is inherited
+            shutil.rmtree(self.tasks_dir)
+        else:
+            logger.warning(
+                "unitxt tasks dir did not start with '%s' and therefor was not deleted",
```
danmcp, line 74: s/therefor/therefore

src/instructlab/eval/unitxt.py
```diff
@@ -0,0 +1,127 @@
...
+                TEMP_DIR_PREFIX,
+            )
+
+    def run(self, server_url: str | None = None) -> tuple:
+        """
+        Runs evaluation
+
+        Returns:
+            overall_scores       Average scores for the task group
```
danmcp, line 83: Misaligned comment

src/instructlab/eval/unitxt.py
```diff
@@ -0,0 +1,127 @@
...
+
+TEMP_DIR_PREFIX = "unitxt_temp"
+
+
+class UnitxtEvaluator(MMLUBranchEvaluator):
+    """
+    An evaluator class, running Unitxt evaluation
+
+    Attributes:
+        model_path      absolute path to or name of a huggingface model
```
danmcp, line 32: nit: s/absolute/Absolute

src/instructlab/eval/unitxt.py
```diff
@@ -0,0 +1,127 @@
...
+
+    name = "unitxt"
+
+    def __init__(
+        self,
+        model_path,
+        unitxt_recipe: str,
+    ):
+        unitxt_task = self.assign_task_name()
+        tasks_dir = self.assign_tasks_dir(unitxt_task)
```
danmcp, line 47: Is this using a local directory?  If so, it needs to be built off a param like output_dir with mt_bench.

danmcp (changes requested):
Thanks for the PR!

src/instructlab/eval/unitxt.py
Roni-Friedman: this is a temporary directory, deleted at the end of the evaluation process. Would you prefer the user specified an output dir? It does not contain anything of use for the user, just the files required for lm eval to run unitxt.

danmcp: I think it would make sense to use the output_dir so it doesn't confuse the user in a local directory.  Also, since you do want to remove at the end, the create/remove logic should probably be in a try/finally block.

src/instructlab/eval/unitxt.py
danmcp: The descriptions weren't aligned top to bottom.  Looks like it's fixed now, thanks!

src/instructlab/eval/unitxt.py
Roni-Friedman: so a user would specify an output dir but will find it doesn't exist at the end of the run?
If the user specifies it, I guess I will not delete it, right?

jwm4: Would it make sense to put all this into a [memory filesystem](https://docs.pyfilesystem.org/en/latest/reference/memoryfs.html)?  In general, it is best to avoid unnecessary disk writes, especially for something that's likely to run in a cloud service where it may or may not have write permissions on some sort of disk.

tests/test_unitxt.py
```diff
@@ -0,0 +1,24 @@
+# First Party
```
danmcp, line 1: Note that with:

https://github.com/instructlab/eval/pull/165

This file should move to scripts/

src/instructlab/eval/mmlu.py
```diff
@@ -153,7 +153,9 @@ def run(self, server_url: str | None = None) -> tuple:
 
         return overall_score, individual_scores
 
-    def _run_mmlu(self, server_url: str | None = None) -> dict:
+    def _run_mmlu(
```
danmcp, line 156: Rather than change the return value here, would you mind making this method return mmlu_output and create a wrapper method called by this class to pull "results" out of it.

src/instructlab/eval/unitxt.py
Roni-Friedman: memory filesystem does not seem to work well, as directory is later accessed also by lm-eval inside the mmlu class and I don't want to start passing this filesystem around (unless owners support such an overall change)

src/instructlab/eval/mmlu.py
Roni-Friedman: if it's ok, I think a wrapper method is not needed, instead I added the following in the mmlu run() method:
for task, result in results **['results']** .items():

Roni-Friedman:
> @Roni-Friedman Could you explain in the description what benefit the Unitxt evaluator would have? Why would a user run the unitxt evaluator over just using the MMLUBranchEvaluator?

Let's discuss this in our meeting as well. I initially wrote a generic evaluator that can use all of unitxt features, but now my understanding is that there are two clear use cases and PR should be adjusted accordingly:
1 - run evaluation on user data
2 - run bluebench

PR #163, opened 2024-10-29
feat: use custom http_client

leseb:
252238b feat: use custom http_client

commit 252238bdd8872d054c0c2cc2ca664d8fd632c47d
Author: Sébastien Han <seb@redhat.com>
Date:   Tue Oct 29 17:56:26 2024 +0100

    feat: use custom http_client
    
    This commit adds the ability to pass a custom HTTP client to the
    MT-Bench evaluator. This is handy when using custom certificates when
    interacting with the judge model serving endpoint.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -10,6 +10,7 @@ dr
 eval
 gpt
 hoc
+http
 instructlab
 jsonl
 justfile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,5 @@
+## 0.4
+
+* Added ability to specify a custom http client to MT-Bench
+
 ## v0.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ accelerate
 pandas
 pandas-stubs
 lm-eval>=0.4.4
+httpx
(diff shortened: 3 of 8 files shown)


nathan-weinberg (approved):
Code LGTM, passing CI, and we are already installing `httpx-0.27.2` so no issues with a new dependency - thanks for the contribution @leseb!

tests/test_branch_gen_answers_with_custom_http_client.py
```diff
@@ -0,0 +1,16 @@
+# Third Party
```
danmcp, line 1: Debating whether this is worth a new file.  These tests still aren't really tests as much as a way to try out the functionality.  I think I would vote for just making this change in the existing "test".

danmcp: My current thinking is we are covering the functional testing pretty well today through the e2e testing.  There are some exceptions so adding functional tests will have some value.  But I would put the priority on adding unit tests.  I think that's the biggest gap and bang for the buck.

danmcp: I started working on the unit/functional test enablement here:

https://github.com/instructlab/eval/pull/165

If we do want to keep this file, note that it should move to scripts/

leseb: Moved the "test" in tests/test_branch_gen_answers.py

danmcp (approved):

PR #165, opened 2024-10-31
Add facilities for unit and functional tests

danmcp:
There was already 1 unit test in place.  The existing functional tests are being moved to scripts for consistency with other projects.

One new functional test has been added to test the mt_bench_branch generator as a starting example

I've included basic tests for mt_bench and mmlu.  Which already found two issues:
- mmlu_branch was using self.tasks instead of the number of results calculating the wrong average
- Not specifying max_workers for mt_bench resulted in an error when it should have defaulted


.github/workflows/test.yml
```diff
@@ -0,0 +1,119 @@
...
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        with:
+          path: ~/.cache/huggingface
+          # config contains DEFAULT_MODEL
+          key: huggingface-${{ hashFiles('src/instructlab/configuration.py') }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install tox tox-gh>=1.2
```
nathan-weinberg, line 103: Should we add `tox-gh` to https://github.com/instructlab/eval/blob/main/requirements-dev.txt and use that file instead?

nathan-weinberg (approved):
Looked over all these changes and honestly everything LGTM - thank you for this @danmcp!

Two nits - one is in the comment, the other would be it'd be nice to add a badge for this new workflow similar to what is done for the CLI here: https://github.com/instructlab/instructlab/blob/main/README.md?plain=1#L4

.github/workflows/test.yml
danmcp: That's probably a good change.  I would lean toward doing it in a separate PR since this version is consistent with the other instructlab repos afaik.  Is that reasonable?

danmcp:
> the other would be it'd be nice to add a badge for this new workflow similar to what is done for the CLI here: https://github.com/instructlab/instructlab/blob/main/README.md?plain=1#L4

Done!

alimaredia (approved):

PR #174, opened 2024-11-07
Add model adapter unit tests

danmcp:


nathan-weinberg (approved):

tests/test_mt_bench_model_adapter.py
```diff
@@ -0,0 +1,61 @@
...
+    (
+        GRANITE_DEFAULT_MODEL_NAME,
+        EXAMPLE_GRANITE_MODEL_PATHS,
+        GraniteAdapter,
+        "ibm-generic",
+    ),
+]
+
+
+def test_get_model_adapter():
```
alimaredia, line 47: Are we going to want to test that a `ValueError` gets thrown if `get_model_adapter()` doesn't have a valid model adapter for the given model path?

danmcp: Added, thanks!

alimaredia (approved):

PR #193, opened 2024-12-06
adds basic ragas eval

RobotSail:
This PR introduces Rubric-based evaluation through Ragas using the default with-reference rubric that they provide. 

The current evaluation supports the two following modes:

1. Being given a `dataset` contains records which hold `user_input` (question), `reference` (golden answer), and `response` (model answer)
2. Being given a `dataset` with the `user_input` and `reference`, and additionally a `ModelConfiguration` of a model which will be used to generate the `response` for each question. This can be any model running on an OpenAI-compatible endpoint

Signed-off-by: Oleg S <97077423+RobotSail@users.noreply.github.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ pandas
 pandas-stubs
 lm-eval>=0.4.4
 httpx
+ragas
--- a/src/instructlab/eval/ragas.py
+++ b/src/instructlab/eval/ragas.py
@@ -5,6 +5,7 @@ from typing import List, Optional, TypedDict
 
 # Third Party
 from langchain_community.chat_models import ChatOpenAI
+from openai import Client as OpenAIClient
 from pandas import DataFrame, read_json
 from pydantic import BaseModel, ConfigDict, field_validator
 from ragas.evaluation import EvaluationDataset, EvaluationResult, RunConfig, evaluate
@@ -16,7 +17,6 @@ from ragas.metrics._domain_specific_rubrics import (  # the rubrics we must inst
 
 # Local
 from .evaluator import Evaluator
-from .mt_bench_common import get_openai_client
 
 
 class Sample(TypedDict):
@@ -49,18 +49,11 @@ DEFAULT_JUDGE_MODEL = "gpt-4o"
 class ModelConfig(BaseModel):
     model_config = ConfigDict(protected_namespaces=())
 
-    # URL of the OpenAI server where the model shall be hosted.
-    base_url: str
-
     # name of the model to use.
     model_name: str
 
     # The system prompt to be used when applying the chat template.
     system_prompt: str = _DEFAULT_SYSTEM_PROMPT
-
-    # We do NOT read from OPENAI_API_KEY for the student model for security reasons (e.g. sending the API key to another client)
-    # To provide an OpenAI key, you must set it here; else the default is used.
-    api_key: str = "no-api-key"
 
     # "model randomness" aka likelihood of sampling something other than the likeliest token
     temperature: float = 0.0
@@ -87,15 +80,18 @@ class RagasEvaluator(Evaluator):
         self,
         student_model: ModelConfig | None = None,
         run_config: RunConfig | None = None,
+        openai_client: OpenAIClient | None = None,
     ):
         self.student_model = student_model
         self.run_config = run_config
+        self.openai_client = openai_client
 
     def run(
         self,
         dataset: List[Sample] | Path,
         student_model: ModelConfig | None = None,
         run_config: RunConfig | None = None,
+        openai_client: OpenAIClient | None = None,
     ) -> EvaluationResult:
         """
         Evaluates the quality of model responses against a graded rubric.
@@ -115,12 +111,16 @@ class RagasEvaluator(Evaluator):
                 a default one is created containing extremely permissive settings when handling
                 timeouts. This is because by default, OpenAI tier-1 usage accounts have very high
                 rate limits resulting in heavy throttling during evaluations.
+            openai_client (openai.Client | None, optional):
+                The client to use when generating questions from the student model, must be compatible with the OpenAI API.
+                This field is required when `student_model` is provided.
 
         Returns:
             EvaluationResult: The results of all evaluations performed by Ragas
         """
         student_model = student_model if student_model else self.student_model
         run_config = run_config if run_config else self.run_config
+        openai_client = openai_client if openai_client else self.openai_client
 
         if not dataset:
             raise ValueError(
@@ -140,14 +140,20 @@ class RagasEvaluator(Evaluator):
         assert input_df is not None
 
         need_to_generate_questions = "response" not in input_df.columns
-        if need_to_generate_questions and not student_model:
+        if need_to_generate_questions and (not student_model or not openai_client):
             raise ValueError(
-                "provided dataset doesn't contain the model `response`, but no `student_model` was provided for inference"
+                "provided dataset doesn't contain the model `response`, but either `student_model` or `openai_client` wasn't provided for inference"
             )
 
         # if the student model was provided then we always generate regardless
         if student_model:
-            input_df = self._generate_answers_from_model(input_df, student_model)
+            if not openai_client:
+                raise ValueError(
+                    "`student_model` was specified but `openai_client` was not provided"
+                )
+            input_df = self._generate_answers_from_model(
+                input_df, student_model, openai_client
+            )
 
         if not run_config:
             # we set extreme timeout/retry values by default since OpenAI tier-1 rate limits
@@ -176,16 +182,15 @@ class RagasEvaluator(Evaluator):
         return results
 
     def _generate_answers_from_model(
-        self, questions: DataFrame, student_model: ModelConfig
+        self,
+        questions: DataFrame,
+        student_model: ModelConfig,
+        openai_client: OpenAIClient,
     ) -> DataFrame:
         """
         Given a DataFrame containing `user_input` columns, generates responses from the given model
         and returns a new DataFrame containing its answers in the `response` column.
         """
-        client = get_openai_client(
-            model_api_base=student_model.base_url, api_key=student_model.api_key
-        )
-
         # initialize response to write into
         updated_df = questions.copy()
         updated_df["response"] = ""
@@ -195,7 +200,7 @@ class RagasEvaluator(Evaluator):
                 student_model.system_prompt,
                 qna["user_input"],
             ]
-            response = client.chat.completions.create(
+            response = openai_client.chat.completions.create(
                 messages=messages,
                 model=student_model.model_name,
                 # specify the seed so we can at least try to have some reproducibility when the clients support it
--- a/src/instructlab/eval/ragas.py
+++ b/src/instructlab/eval/ragas.py
@@ -1,11 +1,15 @@
 # Standard
+from pathlib import Path
 from typing import List, TypedDict
 
 # Third Party
 from langchain_community.chat_models import ChatOpenAI
 from ragas.evaluation import EvaluationDataset, EvaluationResult, RunConfig, evaluate
-from ragas.metrics import RubricsScore
-from ragas.metrics._domain_specific_rubrics import DEFAULT_WITH_REFERENCE_RUBRICS
+from ragas.metrics._domain_specific_rubrics import (  # the rubrics we must instantiate are located inside of a file marked as private
+    DEFAULT_WITH_REFERENCE_RUBRICS,
+    RubricsScore,
+)
+import pandas as pd
 
 # Local
 from .evaluator import Evaluator
@@ -30,13 +34,13 @@ class RagasEvaluator(Evaluator):
         pass
 
     def run(
-        self, dataset: List[Sample], run_config: RunConfig | None = None
+        self, dataset: List[Sample] | Path = None, run_config: RunConfig | None = None
     ) -> EvaluationResult:
         """
         Evaluates the quality of model responses against a graded rubric.
 
         Args:
-            dataset (List[Sample]):
+            dataset (List[Sample] | Path):
                 List of model questions and answers
             run_config (RunConfig | None, optional):
                 Configuration to use when running evaluations. If none is provided, then
@@ -47,6 +51,19 @@ class RagasEvaluator(Evaluator):
         Returns:
             EvaluationResult: The results of all evaluations performed by Ragas
         """
+        if not dataset:
+            raise ValueError(
+                "no dataset was provided, please specify the `dataset` argument"
+            )
+        if isinstance(dataset, Path):
+            input_ds = EvaluationDataset.from_pandas(
+                pd.read_json(dataset, lines=True, orient="records")
+            )
+        elif isinstance(dataset, list):
+            input_ds = EvaluationDataset.from_list(dataset)
+        else:
+            raise TypeError(f"invalid type passed for dataset: {type(dataset)}")
+
         if not run_config:
             # we set extreme timeout/retry values by default since OpenAI tier-1 rate limits
             # are horrible and will result in half of our evaluation results being NaN or 0
@@ -57,10 +74,6 @@ class RagasEvaluator(Evaluator):
                 timeout=3600,
             )
 
-        # we will be using gpt-4o for the foreseeable future, we hardcode this
-        # for consistency of answers
-        input_ds = EvaluationDataset.from_list(dataset)
-
         # default set of metrics
         metrics = [
             RubricsScore(
@@ -68,6 +81,8 @@ class RagasEvaluator(Evaluator):
             )
         ]
 
+        # we will be using gpt-4o for the foreseeable future, we hardcode this
+        # for consistency of answers
         critic_lm = ChatOpenAI(model="gpt-4o")
         results = evaluate(
             dataset=input_ds,
--- a/src/instructlab/eval/ragas.py
+++ b/src/instructlab/eval/ragas.py
@@ -1,3 +1,4 @@
+# # SPDX-License-Identifier: Apache-2.0
 # Standard
 from pathlib import Path
 from typing import List, Optional, TypedDict
@@ -53,7 +54,7 @@ class ModelConfig(BaseModel):
 
     # name of the model to use.
     model_name: str
-    
+
     # The system prompt to be used when applying the chat template.
     system_prompt: str = _DEFAULT_SYSTEM_PROMPT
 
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -10,3 +10,5 @@ pandas
 pandas-stubs
 lm-eval>=0.4.4
 httpx
+
+ragas
```
Maxusmusti, line 14: missing newline at EOF

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class RagasEvaluator(Evaluator):
+    # most basic implementation, we just assume that the user will bring the existing model responses
+    name = "ragas"
+
+    def __init__(self):
+        pass
+
+    def run(
```
Maxusmusti, line 32: So for this a user is expected to bring a list of Sample objects, which hold the input, prediction, and ground truth? Are we going to provide a way to build this list of Samples from given files or lists of each category, or is this moreso just for use with self-built scripts that import the Sample object and build?

RobotSail: Updated it such that `dataset` now is either a `pathlib.Path` object or a `list` of samples, and we read what we need to accordingly

abhi1092:
@RobotSail let me know once you are done with testing the code. Other than that LGTM.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+    # We do NOT read from OPENAI_API_KEY for the student model for security reasons (e.g. sending the API key to another client)
+    # To provide an OpenAI key, you must set it here; else the default is used.
+    api_key: str = "no-api-key"
+
+    # "model randomness" aka likelihood of sampling something other than the likeliest token
+    temperature: float = 0.0
+
+    max_tokens: int = 768
+
+    # Random seed for reproducibility. This is not supported everywhere and therefore is unreliable.
```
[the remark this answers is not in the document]
RobotSail: @alimaredia I'll update this comment because I believe you are confusing it's meaning with our earlier conversation. This comment relates to the seed not being supported by every model serving framework.

RobotSail:
@abhi1092 I've updated the code to have unit tests, please let me know if there was anything I missed.

tests/test_ragas.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # get answers
+        questions = pd.DataFrame({"user_input": ["What is the capital of France?"]})
+        student_model = ModelConfig(
+            base_url="https://api.openai.com",
+            model_name="gpt-3.5-turbo",
+            api_key="test-api-key",
+        )
+        evaluator = RagasEvaluator()
+        result_df = evaluator._generate_answers_from_model(questions, student_model)
```
abhi1092, line 35: @RobotSail shouldn't we hit a mock client here too.

abhi1092: Instead of using gpt-3.5?

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,215 @@
...
+        return results
+
+    def _generate_answers_from_model(
+        self, questions: DataFrame, student_model: ModelConfig
+    ) -> DataFrame:
+        """
+        Given a DataFrame containing `user_input` columns, generates responses from the given model
+        and returns a new DataFrame containing its answers in the `response` column.
+        """
+        client = get_openai_client(
```
abhi1092, line 185: @RobotSail maybe this method and also the class can instead take the client as input?

tests/test_ragas.py
```diff
@@ -0,0 +1,161 @@
...
+            model_name="pt-3.5-turbo",
+            api_key="test-api-key",
+        )
+        run_config = RunConfig(max_retries=3, max_wait=60, seed=42, timeout=30)
+        evaluator = RagasEvaluator()
+
+        ########################################################################
+        # Test case: directly passing a dataset
+        ########################################################################
+        result = evaluator.run(
```
abhi1092, line 133: Here too maybe we can call mock gpt-4o with pre-defined evaluation result. This way we only test the evaluation given the student model response, reference and judge model output

tests/test_ragas.py
RobotSail: @abhi1092 We do test it with a mock-client, GPT-3.5 was just the first model that came to mind as a fill-in.

tests/test_ragas.py
RobotSail: @abhi1092 It is getting mocked, see: https://github.com/instructlab/eval/pull/193/files#diff-07ba3868aa4f9cd9324d6bfebe02ad4b4a3928285c0f01d29ef3d12d3f3b09b4R94

src/instructlab/eval/ragas.py
RobotSail: @abhi1092 That's a good idea. Although the way I've done it here is how it happens in MT-Bench as well. Would it make sense to make a follow-up issue for this?

RobotSail: See: https://github.com/RobotSail/eval/blob/ad12276878dfa691599e37aa5e18d6c1bd7f4afd/src/instructlab/eval/mt_bench_answers.py#L120 for reference

RobotSail: @abhi1092 Actually I will just make this change in this PR since it's small.

RobotSail: Updated the PR to include this.

tests/test_ragas.py
RobotSail: Updated this to use fake values to eliminate the possibility of it actually calling out to the real openai API.

JamesKunstle (commented):
Overall just a could of questions re: function and testing simplification

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+        return v
+
+
+class RagasEvaluator(Evaluator):
+    # most basic implementation, we just assume that the user will bring the existing model responses
+    name = "ragas"
+
+    def __init__(
+        self,
+        student_model: ModelConfig | None = None,
```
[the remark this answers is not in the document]
RobotSail: Yes, because the user isn't required to pass them in at initialization time.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+                This field is required when `student_model` is provided.
+
+        Returns:
+            EvaluationResult: The results of all evaluations performed by Ragas
+        """
+        student_model = student_model if student_model else self.student_model
+        run_config = run_config if run_config else self.run_config
+        openai_client = openai_client if openai_client else self.openai_client
+
+        if not dataset:
```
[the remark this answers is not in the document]
RobotSail: Good point, I'll fix this.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+
+        need_to_generate_questions = "response" not in input_df.columns
+        if need_to_generate_questions and (not student_model or not openai_client):
+            raise ValueError(
+                "provided dataset doesn't contain the model `response`, but either `student_model` or `openai_client` wasn't provided for inference"
+            )
+
+        # if the student model was provided then we always generate regardless
+        if student_model:
+            if not openai_client:
```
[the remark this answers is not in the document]
RobotSail: Good point

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+            student_model: (StudentModelConfig):
+                When this parameter is provided, we'll attempt to use the described model in order to
+                generate the responses from the given list of questions.
+            run_config (RunConfig | None, optional):
+                Configuration to use when running evaluations. If none is provided, then
+                a default one is created containing extremely permissive settings when handling
+                timeouts. This is because by default, OpenAI tier-1 usage accounts have very high
+                rate limits resulting in heavy throttling during evaluations.
+            openai_client (openai.Client | None, optional):
+                The client to use when generating questions from the student model, must be compatible with the OpenAI API.
```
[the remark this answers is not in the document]
RobotSail: This is for the student model. In this PR we are making the opinionated stance that the judge model needs to be GPT-4o for consistent results.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+    model_config = ConfigDict(protected_namespaces=())
+
+    # name of the model to use.
+    model_name: str
+
+    # The system prompt to be used when applying the chat template.
+    system_prompt: str = _DEFAULT_SYSTEM_PROMPT
+
+    # "model randomness" aka likelihood of sampling something other than the likeliest token
+    temperature: float = 0.0
```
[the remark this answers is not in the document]
RobotSail: Thanks for pointing this out, this is way simpler.

src/instructlab/eval/ragas.py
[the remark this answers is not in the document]
JamesKunstle: It might streamline the code if these are required in the constructor so you don't have to do checks below. Although you might be implementing to the standard of the `Evaluator` class, in which case a refactor might make sense in a subsequent PR.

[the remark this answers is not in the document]
RobotSail: I see your point but it'd be fine to leave it as-is. This is also how a lot of other libraries including Ragas implement similar functionality.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        Given a DataFrame containing `user_input` columns, generates responses from the given model
+        and returns a new DataFrame containing its answers in the `response` column.
+        """
+        # initialize response to write into
+        updated_df = questions.copy()
+        updated_df["response"] = ""
+
+        for i, qna in updated_df.iterrows():
+            messages = [
```
cfchase, line 199: Should this be an array of dictionaries `{"role": .., "content": ...}`.  e.g.:

```
messages = [
    {"role": "system", "content": student_model.system_prompt},
    {"role": "user", "content": qna["user_input"]},
]
```

RobotSail: Good point, updating

src/instructlab/eval/ragas.py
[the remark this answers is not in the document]
RobotSail: No actually we can't do this. We have to first compute whether or not we'll need to generate questions, so there's nothing that can be skipped when computing this.

tests/test_ragas.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        ########################################################################
+        # Run the tests
+        ########################################################################
+
+        # Configure all other inputs that Ragas does not depend on for proper mocking
+        student_model = ModelConfig(
+            base_url="https://api.openai.com",
+            model_name="pt-3.5-turbo",
```
alimaredia, line 124: I think this should be `gpt-3.5-turbo`

RobotSail: This is not supposed to be there at all, fixed!

src/instructlab/eval/ragas.py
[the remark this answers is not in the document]
JamesKunstle: That makes sense- could consider renaming api variables so the intention is clear:
`student_model` -> `student_model_config`
`openai_client` -> `student_openai_client`

JamesKunstle (approved):
This seems solid. I have a couple of function naming suggestions, and I think that we could type-safe the incoming data from a .jsonl file to fail as early as possible.

src/instructlab/eval/ragas.py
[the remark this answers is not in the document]
JamesKunstle: So the user doesn't accidentally create a client to 4o-mini thinking that they literally need a client pointing to ChatGPT.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+                seed=DEFAULT_SEED,
+                timeout=3600,
+            )
+
+        metrics = self._get_metrics()
+        evaluation_ds = EvaluationDataset.from_pandas(input_df)
+
+        # we will be using gpt-4o for the foreseeable future, we hardcode this
+        # for consistency of answers
+        critic_lm = ChatOpenAI(model=DEFAULT_JUDGE_MODEL)
```
alimaredia, line 162: How is the API key supposed to be passed into the judge model? If we're assuming the environment variable is set for that we need  When I run the script you the short script you sent me yesterday I hit that as this as an issue.

There were previous iterations of this PR that had a `base_url` and a key in the `ModelConfig` that I think are needed since the student and the judge need at least a `base_url` if not both.

src/instructlab/eval/ragas.py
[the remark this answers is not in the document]
RobotSail: Yes I think that's a good suggestion now that we've decoupled the OpenAI config from being in the student model. Good catch!

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+                max_retries=120,
+                max_wait=7200,
+                seed=DEFAULT_SEED,
+                timeout=3600,
+            )
+
+        metrics = self._get_metrics()
+        evaluation_ds = EvaluationDataset.from_pandas(input_df)
+
+        # we will be using gpt-4o for the foreseeable future, we hardcode this
```
alimaredia, line 160: Could we still make it possible to pass in a model but set 4o as the default judge model? In my development for the ilab interface I'm using 4o-mini just because it's cheaper and it would be nice to have that flexibility as was in previous iterations.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+        if isinstance(dataset, list):
+            input_df = DataFrame(dataset)
+        elif isinstance(dataset, Path):
+            input_df = read_json(dataset, orient="records", lines=True)
+        else:
+            raise TypeError(f"invalid type of dataset: {type(dataset)}")
+
+        # this should never happen, but pylint is not smart enough to detect it
+        assert input_df is not None
+
```
alimaredia, line 130: @RobotSail what do you think about adding an `assert` here if `user_input` and `reference` aren't in the `input_df`, just to make sure the `.jsonl` is in the right format.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+        need_to_generate_questions = "response" not in input_df.columns
```
alimaredia, line 131: This could be future work, but if we're doing any logging it would be great to log, say at debug level, when this bit gets set.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+        if need_to_generate_questions and (not student_model or not openai_client):
+            raise ValueError(
+                "provided dataset doesn't contain the model `response`, but either `student_model` or `openai_client` wasn't provided for inference"
+            )
+
+        # if the student model was provided then we always generate regardless
+        if student_model:
```
alimaredia, line 138: Can we combine this conditional and the one above it. It seems if `needs_to_generate_questions` is set then we'd run `_generate_answers_from_model()`.

The reason I say that is I accidentally passed in input with responses, but also the student model and the openai_client when I was using this PR. What would you want to happen in that case? Ignore the responses column and ask the student model the questions and get the responses?

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+                model=student_model.model_name,
+                # specify the seed so we can at least try to have some reproducibility when the clients support it
+                seed=42,
+                max_tokens=student_model.max_tokens,
+                temperature=student_model.temperature,
+            )
+            updated_df.at[i, "response"] = response.choices[0].message.content
+        return updated_df
+
+    def _get_metrics(self) -> List[Metric]:
```
alimaredia, line 206: What's the use of this function? Since it's only being used in one place and we can't configure what's being passed into `rubrics`.

src/instructlab/eval/ragas.py
RobotSail: I think since we want to lock it down to OpenAI for now just to have a consistent evaluation base, we can just expose the API key. We can make a follow-up where this is something configurable.

src/instructlab/eval/ragas.py
```diff
@@ -0,0 +1,212 @@
...
+        student_model = student_model if student_model else self.student_model
+        run_config = run_config if run_config else self.run_config
+        openai_client = openai_client if openai_client else self.openai_client
+
+        # ensure we are in the dataframe format
+        input_df = None
+        if isinstance(dataset, list):
+            input_df = DataFrame(dataset)
+        elif isinstance(dataset, Path):
+            input_df = read_json(dataset, orient="records", lines=True)
```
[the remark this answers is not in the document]
RobotSail: Sure, I don't see a reason not to.

src/instructlab/eval/ragas.py
RobotSail: Done

src/instructlab/eval/ragas.py
RobotSail: Done

src/instructlab/eval/ragas.py
RobotSail: Yeah the idea is that when the student model is provided then it will automatically be used to generate responses. But the reason why we can't combine these two conditionals is that there's a separate state where the student model was not passed in, but there isn't sufficient information to perform the evaluation (e.g. the `response` field is missing), then we error out because we cannot continue.

RobotSail: I do think that this can be designed better, but let's focus on getting this PR merged. Then we can focus on improvements afterwards.

src/instructlab/eval/ragas.py
RobotSail: Sure

src/instructlab/eval/ragas.py
RobotSail: The point is that since we select the metrics carefully, having them isolated allows us to treat them with more care. Also - when you add to this list it becomes very messy, so it's just slightly more readable to have it in a separate method.

src/instructlab/eval/ragas.py
alimaredia: Agree we can always tweak later

src/instructlab/eval/ragas.py
RobotSail: This comment is outdated, we've since updated this evaluation to also accept the judge model name as well.

alimaredia (approved):

PR #211, opened 2025-01-12
ragas: pass rubric in via RubricScore.SingleTurnPrompt.instruction

alimaredia:
In ragas, RubricScores.rubrics isn't being used anywhere except in __repr__ thus it was not being passed into the prompt for the judge to evaluate responses against reference answers.

Passing a string rubric into SingleTurnPrompt.instruction allows us to pass the rubric into the prompt sent
to the judge.

### DIFF
--- a/src/instructlab/eval/ragas.py
+++ b/src/instructlab/eval/ragas.py
@@ -1,20 +1,24 @@
 # # SPDX-License-Identifier: Apache-2.0
 # Standard
-from pathlib import Path
 from typing import TYPE_CHECKING, List, Optional, TypedDict
+import os
 
 # Third Party
 from langchain_community.chat_models import ChatOpenAI
 from openai import Client as OpenAIClient
+from openai import OpenAI, OpenAIError
 from openai.types.chat import ChatCompletionMessageParam
-from pandas import DataFrame, read_json
+from pandas import DataFrame
 from pydantic import BaseModel, ConfigDict, Field
 from ragas.evaluation import EvaluationDataset, EvaluationResult, RunConfig, evaluate
 from ragas.metrics import Metric
 from ragas.metrics._domain_specific_rubrics import (  # the rubrics we must instantiate are located inside of a file marked as private
     RubricsScore,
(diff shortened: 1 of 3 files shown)


alimaredia:
@RobotSail Here is what the prompts and internal responses look like at without this PR and with this PR:
https://paste.sh/KB9OjHDP#29Fd_qd5QQqCoaV7zMtm6NSG

What should the final prompt look like? There's a lot baked by ragas that's easy to see in the output when this PR is not applied.

src/instructlab/eval/ragas.py
```diff
@@ -88,29 +96,25 @@ def __init__(
...
             raise ValueError(
-                f"invalid dataset provided, missing the following keys: {', '.join(missing_keys)}"
+                f"Dataset can only have the following keys: {', '.join(required_keys)}. Keys provided were: {', '.join(columns_list)}"
             )
 
     def run(
         self,
-        dataset: List[Sample] | Path,
-        student_model: ModelConfig | None = None,
+        dataset: List[Sample] | DataFrame,
```
RobotSail, line 116: How come we are getting rid of the `Path` type here?

src/instructlab/eval/ragas.py
```diff
@@ -150,50 +149,29 @@ def run(
         judge_openai_api_key = (
             judge_openai_api_key if judge_openai_api_key else self.judge_openai_api_key
         )
-        student_model = student_model if student_model else self.student_model
         run_config = run_config if run_config else self.run_config
-        student_openai_client = (
```
RobotSail, line 155: Why are we deleting this??

RobotSail (changes requested):
Thanks for this PR @alimaredia.

I'm a little confused by this PR, because it seems like there are a lot of things being changed which may not be intended to be. 

Could you please update this PR to only be changing the intended behavior (making sure that the old ragas rubric template can still be used)? 

Additionally, we should somehow make the old rubric usage be a toggle-able behavior, since there is reason for which the users may want to use the new rubric.

PR #212, opened 2025-01-18
Bump ragas version

alimaredia:
Before ragas v0.2.11 RubricScores.rubrics wasn't being applied properly. This commit sets that as the minimum version for this library.

A change in v0.2.11 from previous versions was a change in the prompt for domain specific knowledge evaluation with reference. The prompt from previous versions is now explicitly passed in.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,4 @@ pandas
 pandas-stubs
 lm-eval>=0.4.4
 httpx
-ragas
+ragas>=0.2.11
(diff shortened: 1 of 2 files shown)


RobotSail (commented):
Thanks for this PR @alimaredia. I feel like we shouldn't _force_ the old template to be used. Maybe we should allow the legacy template to be used with some kind of legacy flag?

RobotSail (approved):
LGTM

PR #234, opened 2025-03-17
Implement leaderboard as a benchmark

RobotSail:
This PR contributes the Open LLM Leaderboard v2 to become an evaluation exposed within instructlab/eval.

In particular, this exposes leaderboard with the ability for users to select a subset of the tasks in leaderboard.

In addition, this benchmark is implemented in a way such that it runs each subtask on the most optimal inference backend for a given task.

Specifically, MCQ-style tasks (GPQA, MUSR, MMLU-Pro, and BBH) are executed directly through regular HF transformers, whereas generative tasks (IFEval and MATH-Hard) get executed through vLLM.


bbrowning (approved):
Assuming, the failing unit test around `StrEnum` import is figured out, this looks good. The new code is purely additive, and the extra requirements are in a separate requirements txt file so it shouldn't cause any issues with downstream builds.

PR #250, opened 2025-05-02
test run

jaideepr97:


src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def __init__(
+        self,
+        model_path: Optional[str] = None,
+        output_file: Optional[str] = None,
+        tasks: list[str] = RULER_TASKS,
+        num_gpus: Optional[int] = None,
+        eval_config: Optional[Dict[str, Any]] = None,
+        vllm_config: Optional[Dict[str, Any]] = None,
+        hf_config: Optional[Dict[str, Any]] = None,
```
RobotSail, line 43: I don't think you need these since you're only calling this through a local OpenAI endpoint

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    Class definition for running RULER benchmarking tasks.
+    """
+
+    def __init__(
+        self,
+        model_path: Optional[str] = None,
+        output_file: Optional[str] = None,
+        tasks: list[str] = RULER_TASKS,
+        num_gpus: Optional[int] = None,
```
RobotSail, line 40: You can drop this too

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        eval_config: Optional[Dict[str, Any]] = None,
+        vllm_config: Optional[Dict[str, Any]] = None,
+        hf_config: Optional[Dict[str, Any]] = None,
+        openai_config: Optional[Dict[str, Any]] = None,
```
RobotSail, line 44: I see that you've declared this as an argument and are reading it further down, but I don't see it actually being passed into lm-eval-harness. You'll want to make sure we're doing this by following how I did it here: https://github.com/instructlab/eval/blob/f48c6a1833d98e0bb3cf8931e587ce5102c686a6/src/instructlab/eval/leaderboard.py#L745

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Store evaluation configurations
+        self.eval_config = eval_config or {}
+        self.vllm_config = vllm_config or {}
+        self.hf_config = hf_config or {}
+        self.openai_config = openai_config or {}
+
+        self.api_endpoint = api_endpoint or None
+        self.num_gpus = num_gpus
+        self.max_length = max_length or 4096
```
RobotSail, line 61: 4096 here seems arbitrary, I would move it up to be a top-level constant and include a comment explaining why you're picking this number as a default (it's the lm-eval-harness default I'm pretty sure).

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        max_length: Optional[int] = 4096,
+    ) -> None:
+        """
+        Run the RULER evaluation using the specified model and tasks.
+        """
+
+        model_path = self.model_path if model_path is None else model_path
+        tasks = self.tasks if not tasks else tasks
+        output_file = self.output_file if not output_file else output_file
+
```
RobotSail, line 140: We probably want to add a validation step here before we run through `simple_evaluate` to guarantee that the `model_path`, `tasks` and any other variables we've defined above are for sure `not None` (and any other validations that we would ∑ant to check, like ensuring `output_file` can be written to).

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        api_endpoint: Optional[str] = None,
+        max_length: Optional[int] = None,
+    ) -> None:
+        self.model_path = model_path
+        self.tasks = tasks
+        self.results: Dict[Any, Any] = {}
+        self.output_file = output_file
+
+        # Store evaluation configurations
+        self.eval_config = eval_config or {}
```
RobotSail, line 54: It doesn't seem like this is getting read anywhere. Should `lm-eval-harness` read this when we call `simple_evaluate`?

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        self.hf_config = hf_config or {}
+        self.openai_config = openai_config or {}
+
+        self.api_endpoint = api_endpoint or None
+        self.num_gpus = num_gpus
+        self.max_length = max_length or 4096
+
+    def save_to_file(self, output_file: Optional[str] = None) -> None:
+        """Save results to a JSON file"""
+        output_file = output_file or self.output_file
```
RobotSail, line 65: Always prefer to use this syntax and leave `or` as only for boolean logic:

```suggestion
        output_file = output_file if output_file else self.output_file
```

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Process the evaluation results from lm_eval for the given file path and extract
+        aggregarted scores for each context length
+        Args:
+            fpath (pathlib.Path): The file path to the evaluation results.
+
+        """
+        unqiue_metrics_dict: dict[str, Any] = {}
+
+        def extract_metrics(results: dict, unqiue_metrics_dict: dict = {}):
```
RobotSail, line 87: You may want to add a comment here explaining why we need to do this so anyone who's unfamiliar with lm-eval-harness or even this particular benchmark would have an idea of what this solves. It would be good even to provide an example of the schema that we're dealing with just as a code comment.

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        if fpath:
+            with open(fpath, "r", encoding="utf-8") as f:
+                raw_results = json.load(f)
+
+        if raw_results is not None:
+            extract_metrics(raw_results["results"], unqiue_metrics_dict)
+        unique_float_metrics = {}
+        # if value is list of floats, average the list
+        for k, v in unqiue_metrics_dict.items():
+            if isinstance(v, list) and all(isinstance(i, float) for i in v):
```
RobotSail, line 109: Nice I like this

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+            v for v in unique_float_metrics.values() if isinstance(v, float)
+        ]
+        if float_values:
+            unique_float_metrics["avg"] = sum(float_values) / len(float_values)
+        else:
+            unique_float_metrics["avg"] = 0.0
+
+        # result format
+        # {'8192': 0.90, '32768': 0.82, '65536': 0.77, '131072': 0.71, 'avg': 0.80}
+        self.results = unique_float_metrics
```
RobotSail, line 123: @jaideepr97 Would it make sense for this method to return this value instead of storing it? And then the caller can decide to store it on the object if they want? I.e. this can just be made into a static class method I feel like.

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+            "max_length": max_length,
+        }
+
+        lm_eval_results = simple_evaluate(
+            model="local-completions",
+            model_args=model_args,
+            tasks=tasks,
+        )
+
+        self.process_lm_eval_results(
```
RobotSail, line 154: Here I would just get a value returned from this method, store it on the object and then write it to disk.

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,161 @@
...
+        output_file = self.output_file if not output_file else output_file
+
+        # Prepare model_args
+        model_args = {
+            "pretrained": model_path,
+            "base_url": api_endpoint,
+            "max_length": max_length,
+        }
+
+        lm_eval_results = simple_evaluate(
```
RobotSail, line 148: Might be worthwhile to store the raw result from lm-eval-harness somewhere. While there are cleaner solutions, I think it would suffice to store it on a `._result` field.

RobotSail (commented):
Thank you for the PR @jaideepr97 ! I left some feedback but overall this looks like it's in a pretty good shape.

src/instructlab/eval/ruler.py
jaideepr97: this was from when i was trying to follow the same structure as other evaluators and was looking into what might be needed to support vllm and hf backend as well

since we agreed to keep this PR just to local openai, ive gone ahead and removed some of the unused fields like this one, which can be added in a follow up PR as needed

src/instructlab/eval/ruler.py
jaideepr97: fixed, just raising errors for now

src/instructlab/eval/ruler.py
jaideepr97: updated

src/instructlab/eval/ruler.py
jaideepr97: got rid of these for now, since they werent being used

src/instructlab/eval/ruler.py
```diff
@@ -0,0 +1,169 @@
...
+    Class definition for running RULER benchmarking tasks.
+    """
+
+    name = "ruler"
+
+    def __init__(
+        self,
+        model_path: Optional[str] = None,
+        output_file: Optional[str] = None,
+        tasks: list[str] = RULER_TASKS,
```
RobotSail, line 43: In Python, you want to avoid using complex objects like Lists, Dicts, Sets, etc. as defaults when declaring functions. The reason is that you are signing the default as a pointer to the instantiated object (in this case `RULER_TASKS`). So the workaround we do is to have a function setup like:

```py
def foo(my_list: list = None):
  if not my_list:
    my_list = []
```

In the case of this `__init__` method, since `RULER_TASKS` is a list which already contains data, we would want to create a copy of it for this default, which we can do with the slice notation (`[:]`):

```py
def __init__(self, tasks: list[str] = None):
  if not tasks:
    # create copies of the strings in RULER_TASKS
    self.tasks = RULER_TASKS[:]
  else:
    self.tasks = tasks

```

src/instructlab/eval/ruler.py
RobotSail: I see the constant defined above but it doesn't seem like you're using it

src/instructlab/eval/ruler.py
RobotSail: Nice, this is perfect. Thank you @jaideepr97 !

RobotSail (approved):
LGTM!

PR #255, opened 2025-05-15
deps: Bump minimum required Python version to 3.11

courtneypacheco:
## Overview

In our core repo, we require Python >=3.11: https://github.com/instructlab/instructlab/blob/main/pyproject.toml#L13

In this repo, however, we currently require Python >=3.9, and we are only testing against Python 3.10 and 3.11 in the CI. Therefore, we should ensure consistency in terms of Python version support throughout our GitHub org.

## Proposed Changes
This bump includes updates to documentation, Git workflows, test configs, etc.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - "3.10"
           - "3.11"
         platform:
           - "ubuntu-latest"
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -45,7 +45,6 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - "3.10"
```
courtneypacheco, line 48: Note: We ultimately want to add Python 3.12 support, which is why I want to leave the matrix logic as is despite us only testing with 3.11 right now.

RobotSail (approved):

booxter (approved):

booxter:
@mergifyio rebase

PR #260, opened 2025-05-20
ci: Fix e2e job failures

booxter:
- **fix: Require medium and large E2E jobs to use constraints-dev.txt to unblock jobs**
- **ci: Install flash-attn using the same torch version before reqs**


booxter:
As @courtneypacheco correctly points out, this change applies to pull_request_target workflow, meaning it won't be tested in the currently running CI. To test it, we would have to push the branch to main repo, then manually start a workflow of interest using this branch. I don't have write access to the repo, so I can't do it even if I wanted. Besides, I think it's crucial that eval folks are involved in fixing the CI for the repo. I am an outsider and can only guide.

RobotSail (approved):
Thanks Ihar! LGTM :shipit:

